PR #16, opened 2018-06-22
Fix kill-telega message.

sarg:


telega.el
```diff
@@ -60,9 +60,13 @@
...
-  (when (or force (y-or-n-p (format "Kill telega (and all %d chat buffers)? "
-                                    (length telega--chat-buffers))))
-    (kill-buffer telega-root-buffer-name)))
+  (let ((chat-count (length telega--chat-buffers)))
+    (when (or force (y-or-n-p (concat "Kill telega"
+                                      (cond ((eq chat-count 0) "")
+                                            ((eq chat-count 1) (format " (and 1 chat buffer)"))
+                                            ('t (format " (and all %d chat buffers)" chat-count)))
+                                      "? ")))
+              (kill-buffer telega-root-buffer-name))))
```
zevlg: `t` does not require quotation

telega.el
```diff
@@ -60,9 +60,13 @@
   "Kill currently running telega.
 With prefix arg force quit without confirmation."
   (interactive "P")
-  (when (or force (y-or-n-p (format "Kill telega (and all %d chat buffers)? "
-                                    (length telega--chat-buffers))))
-    (kill-buffer telega-root-buffer-name)))
+  (let ((chat-count (length telega--chat-buffers)))
```
zevlg: Best practice is to have bindings as late as you can.  In this case `chat-count` binds, but if `force` is `non-nil`, then `chat-count` is not used

PR #17, opened 2018-06-22
Implement photo preview.

sarg:
Works for me on emacs 27.
Photos load on the fly. The solution is not very polished, so feedback is welcome.


sarg:
Weird commit list after rebase. I'll fix it and open another PR.

PR #20, opened 2018-06-25
Feature/stickers

sarg:


sarg:
Seems that when ImageMagick is built with WebP support we can omit preview conversion

PR #23, opened 2018-06-27
Don't kill already dead telega.

sarg:
When telega isn't running, `telega-kill` produces an error.


sarg:
I have force-pushed fix.

PR #30, opened 2018-07-07
Allow user to apply markdown formatting implicitly.

sarg:


telega-chat.el
```diff
@@ -940,9 +940,14 @@ Pass non-nil FROM-BACKGROUND if message sent from background."
...
-(defun telega-chat-send (markdown)
+(defcustom telega-chat-use-markdown-formatting nil
+  "*Use markdown formatting for outgoing messages."
+  :type 'boolean
+  :group 'telega)
+
+(defun telega-chat-send (prefix-arg)
   "Send current input to the chat.
-With prefix arg, apply markdown formatter to message."
+With PREFIX-ARG, apply markdown formatter to message."
```
furiel: Can you explain this help text? Me seems incomplete. According to the patch, if I set `telega-chat-use-markdown-formatting` to true, and I use prefix-arg, then markdown formatting is disabled. What is the intended behavior here? Toggling the value of `telega-chat-use-markdown-formatting`?

sarg: Yeah, the text is misleading. I should have changed it. The idea is that prefix-arg inverts your default setting.

furiel:
I earlier had a bad experience with markdown formatting in telegram. This is not telega specific, but telegram in general. Same goes with html formatting.

The problem is if someone tries to send a malformed markdown message, then the telegram server emits a parse error, and does not send the message, the message is lost. One can try this by sending `_test` with markdown enabled in telega, and receive similar error.
```
telega-server error: Can't parse entities: Can't find end of the entity starting at byte offset 0
telega-server error: Message text can't be empty
```

Same phenomenon is reproducible without tdlib, with curl.
```
$ curl --data-ascii "chat_id=..." --data-ascii "parse_mode=markdown" --data-ascii "text=_hello" https://api.telegram.org/bot[...]/sendMessage 
{"ok":false,"error_code":400,"description":"Bad Request: can't parse entities: Can't find end of the entity starting at byte offset 0"}
```
As an enhancement (feature request?): if possible, telega could try to validate markdown, and in case of failure, falls back to parse_mode=None

furiel:
Btw `_test` seems an extreme case, but these do not work either: `5*5=25` or `telega_server`

zevlg:
@furiel message is not lost actually, you can pop it up from history using `M-p` binding, edit and resend.

Having some fallback in case of invalid markdown is reasonable

furiel:
> @furiel message is not lost actually, you can pop it up from history using M-p binding, edit and resend.

Thanks for the hint. That's neat.

telega-chat.el
```diff
@@ -947,7 +947,7 @@ Pass non-nil FROM-BACKGROUND if message sent from background."
 
 (defun telega-chat-send (prefix-arg)
   "Send current input to the chat.
-With PREFIX-ARG, apply markdown formatter to message."
+With PREFIX-ARG, inverses your current preference for markdown formatting."
```
zevlg: Add please ref to `telega-chat-use-markdown-formatting` into docstring, so it is clear what you mean by "preference"

telega-chat.el
```diff
@@ -962,7 +967,10 @@ With prefix arg, apply markdown formatter to message."
     (setq telega-chatbuf--input-idx nil
           telega-chatbuf--input-pending-p nil)
 
-    (apply sfunc telega-chatbuf--chat input markdown sargs)))
+    (apply sfunc telega-chatbuf--chat input
+           (or (and (not prefix-arg) telega-chat-use-markdown-formatting)
+               (and prefix-arg (not telega-chat-use-markdown-formatting)))
+           sargs)))
```
zevlg: This is hard to comprehend, could you please rewrite it just to mean "reverse markdown preference", like:

```elisp
(if telega-chat-use-markdown-formatting
   (not prefix-arg)
 prefix-arg)
```

telega-chat.el
```diff
@@ -940,9 +940,14 @@ Pass non-nil FROM-BACKGROUND if message sent from background."
   (interactive)
   (telega-chat-buffer--prompt-reset))
 
-(defun telega-chat-send (markdown)
+(defcustom telega-chat-use-markdown-formatting nil
+  "*Use markdown formatting for outgoing messages."
```
zevlg: Gerenal rule for writing docstring for boolean vars is to start it by "*Non-nil to XXX" or "For non-nil, XXX", where XXX action you take for non-nil value.  In our case it should be: "*Non-nil to use markdown formatting for outgoing messages."

zevlg:
Nice feature, thanks @sarg !

furiel:
@zevlg thanks for the quick implementation! I just tried and it works indeed.

PR #38, opened 2018-10-13
telega.el: added guard for "print-circle" variable

wiedzmin:
We should disable linking sexp substructures, because
it breaks parsing on server side.


zevlg:
Nice catch, thanks!

wiedzmin:
Thanks for the great tool!

PR #55, opened 2019-04-10
Imporve the logic of getting filename from uri under windows system.

cireu:
As the title

Ported from https://github.com/joaotavora/eglot/blob/master/eglot.el#L1032

Warning: This commit haven't been tested on Windows system!


zevlg:
Thanks for the PR,  we will test it on win32 first and then apply with minor changes

Issue #73, opened 2019-08-06
Use `define-minor-mode` to create `telega-notifications-mode`


cireu:
`defin-minor-mode` is emacs builtin minor mode management mechanism, it supports
interactive toggling, and it will create hooks for minor-mode. 

In my option, using `define-minor-mode` is better than adhoc entry function.

If you agree with me, I'll open a PR to do this.

zevlg:
Minor mode is something buffer handling specific. Notifications is more like tracking, just toggles some inner telega functionality.  It is also good example how telega hooks can be used

cireu:
On Tue, 06 Aug 2019 16:07:44 +0800,
Zajcev Evgeny wrote:
> 
> Minor mode is something buffer handling specific. Notifications is more like tracking, just toggles some inner telega functionality. It is also good example how telega
> hooks can be used
> 
> ―
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or mute the thread.

Minor mode can also be globalized, you need to specify the `:global` argument to
t

> :global GLOBAL	If non-nil specifies that the minor mode is not meant to be
> 		buffer-local, so don't make the variable MODE buffer-local.
> 		By default, the mode is buffer-local.

cireu:
At least, allow customize(I mean, using `M-x customize`) `telega-use-notifications` after loaded `telega` can effect current telega instance. You can define a setter for `telega-use-notifications` with `:set`

zevlg:
> At least, allow customize(I mean, using `M-x customize`) `telega-use-notifications` after loaded `telega` can effect current telega instance. You can define a setter for `telega-use-notifications` with `:set`

yeah, that is reasonable, I'll do this

zevlg:
@cireu I just realized, that there is nothing wrong at all making notifications to be global minor mode.

There will be more minor modes for telega, so why not make notifications same way

cireu:
Submit a PR to do this.

PR #75, opened 2019-08-16
[enh] Make `telega-notifications-mode` a minor mode.

cireu:
https://github.com/zevlg/telega.el/issues/73


zevlg:
merged, thanks!  I think we can remove `telega-use-notifications` then

PR #89, opened 2019-09-13
change pipe2 to pipe

SteamedFish:
when flag is 0, pipe2 is same as pipe

macOS don't have pipe2, use pipe can make it compatiple with macOS.


zevlg:
Nice catch, thanks!

PR #95, opened 2019-10-20
[fix] Force utf-8 while starting telega server.

cireu:
If user change default coding system, telega may failed to send non-ASCII
message and core dumped. This PR can fix this.

How to reproduce:

-  `(set-default-coding-systems 'gbk)`, or any other except utf-8 series.
- Start telega and send any message contains non-ASCII character.
- telega-server core dumped.

----

#


zevlg:
Thanks, will simple
```elisp
(set-process-coding-system proc 'utf-8 'utf-8)
```
do the job?

cireu:
Yes, it also works.

If you sure there's no BOM in data. You can simply use utf-8 instead of utf-8-auto.

zevlg:
Merged in, thanks!

PR #98, opened 2019-10-24
Tty compatibility

akater:
Telega is barely usable in tty when emacs is bult `--without-librsvg`. The patch fixes this by wrapping all image-related functions in display-graphic-p checks.

Hopefully, imagemagick and librsvg won't be hard dependencies in the end but I'm not there yet. The proposed patches help a lot already, though.

### DIFF
--- a/telega-chat.el
+++ b/telega-chat.el
@@ -2545,8 +2545,11 @@ If PREVIEW-P is non-nil, then generate preview image."
   "Attach FILENAME as photo to the current input."
   (interactive (list (read-file-name "Photo: ")))
   (let ((ifile (telega-chatbuf--gen-input-file filename 'Photo t))
-        (img-size (image-size
-                   (create-image filename 'imagemagick nil :scale 1.0) t)))
+        (img-size
+         (if (display-graphic-p)
+             (image-size
+              (create-image filename 'imagemagick nil :scale 1.0) t)
+           (cons 0 0))))
     (telega-chatbuf-input-insert
      (list :@type "inputMessagePhoto"
            :photo ifile
--- a/telega-media.el
+++ b/telega-media.el
@@ -572,43 +572,45 @@ File is specified with FILE-SPEC."
          (name (if (eq (telega--tl-type chat-or-user) 'user)
                    (telega-user--name chat-or-user)
                  (telega-chat-title chat-or-user))))
-    (if (telega-file-exists-p photofile)
-        (let ((img-type (image-type-from-file-name photofile))
-              (clip (telega-svg-clip-path svg "clip")))
-          (svg-circle clip (/ xw 2) (/ cfull 2) (/ ch 2))
-          (svg-embed svg photofile
-                     (format "image/%S" img-type)
-                     nil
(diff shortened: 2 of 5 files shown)


cireu:
Hello! I suggest you wrap `create-image` and `svg-image` into a new function like `telega-create-image` and move the if-else branch inside the new function. It can reduce tons of repeat `(if (display-graphic-p) ...)`.

akater:
I actually wanted to discuss that. I believe a macro might be more appropriate here. I suggest a macro that captures `IMAGE-PROPERTIES`; definition and examples follow. If you're fine with this, I'll introduce it right away. There'd have to be two functions, for `create-image` and `svg-image`, or a single but bloated one. You also often have arbitrary code around that uses image-properties list. Given all this, I think a macro is appropriate. Knowing that some people might object to variable capture, I wanted to ask first.

If you don't approve of this, I'll implement `telega-create-image` and `telega-svg-image` or some `telega-make-image` that encompasses both. I'm afraid this will be ugly.

All modifications made by now were made with this macro in mind.

# Definition
```emacs-lisp
(defmacro with-tty-safe-image-properties (props-form &rest body)
  "With IMAGE-PROPERTIES variable captured and set to PROPS-FORM evaluated, evaluate BODY in graphics-enabled frames, and return ('dummy-image . IMAGE-PROPERTIES) list otherwise."
  `(let ((image-properties ,props-form))
     (if (display-graphic-p)
         (progn ,@body)
       (cons 'dummy-image image-properties))))
```

# Examples
## Insert image conditionally
```emacs-lisp
(macroexpand
 '(with-tty-safe-image-properties
   (list :scale 1.0
         :width 42 :height 42
         :ascent 'center
         :telega-text (make-string 6 ?X))
   
   (snip)
   (apply #'create-image 'img 'imagemagick nil
          image-properties)))
#+end_src

=>
```emacs-lisp
(let
    ((image-properties
      (list :scale 1.0 :width 42 :height 42 :ascent 'center :telega-text
            (make-string 6 88))))
  (if
      (display-graphic-p)
      (progn
        (snip)
        (apply #'create-image 'img 'imagemagick nil image-properties))
    (cons 'dummy-image image-properties)))
```

## Always insert dummy-image in non-graphic terminals
```emacs-lisp
(flet ((display-graphic-p nil nil))
  (with-tty-safe-image-properties
   (list :scale 1.0
         :width 42 :height 42
         :ascent 'center
         :telega-text (make-string 6 ?X))
   
   (snip)
   (apply #'create-image 'img 'imagemagick nil
          image-properties)))
```

=>
```emacs-lisp
(dummy-image :scale 1.0 :width 42 :height 42 :ascent center :telega-text "XXXXXX")
```

jellelicht:
Is this still somehow being worked on, or has it been abandoned (for now)?

zevlg:
This patch is bad in its core, you can't rely logic on `display-graphic-p`, because this is a frame property.  And telega buffers might be displayed in different frames at the same time.

zevlg:
> Is this still somehow being worked on, or has it been abandoned (for now)?

JFYI: emacs-nox without svg support can run telega - see #219

PR #102, opened 2019-11-22
update current status on emacs-mac

SteamedFish:
Emacs-mac has added a new compiling option to compile emacs to use rsvg
instead of webkit to render svg files, witch fixed the hangup issue.

But there's still one issue exist with emacs-mac: stickers cannot be
displayed correctly.


zevlg:
Merged in, thanks!

Issue #72, opened 2019-08-05
Is imagemagick required for emacs27?


emacsomancer:
Emacs 27 builds by default without imagemagick, using its own internal libraries to handle graphics. Can this be used by telega?

cireu:
Can you point out what's new in image manipulation API of Emacs 27?

emacsomancer:
@cireu 

From Emacs 27 NEWS:

> ** Emacs no longer defaults to using ImageMagick to display images,
due to security and stability concerns. 
.....
> +++
> ** Emacs now supports resizing and rotating images without ImageMagick.
All modern systems support this feature.  (On GNU and Unix systems,
Cairo drawing or the XRender extension to X11 is required for this to
be available; the configure script will test for it and, if found,
enable scaling.)

> The new function 'image-transforms-p' can be used to test whether any
given frame supports these capabilities.

(Searching in Emacs NEWS for 'imagemagick' provides a  few more details

zevlg:
Imagemagick is no longer required.

PR #106, opened 2019-11-28
Improve image handling when imagemagick is not available

alexmurray:


telega-customize.el
```diff
@@ -129,7 +129,7 @@ Make sure you have tracking.el loaded if this option is enabled."
   :type 'boolean
   :group 'telega
   :set (lambda (option value)
-         (if (fboundp 'imagemagick-types)
+         (if (or (fboundp 'imagemagick-types) (version< "27.0.49" emacs-version))
              (set-default option value)
```
zevlg: It is better to use `(and (fboundp 'image-transforms-p) (funcall 'image-transforms-p))` instead of Emacs version comparison.

zevlg:
https://github.com/zevlg/telega.el/issues/72 will require more work, however this is patch is good as first start to support built-in images transforms in Emacs27.

Please see my comment in the code, and I'll merge

Thanks

zevlg:
Merged in, thanks!

Partly fixes https://github.com/zevlg/telega.el/issues/72

But still, support for stickers is required, in case when imagemagick is not available

zevlg:
JFYI: Stickers now supported in non-imagemagick setup

PR #111, opened 2019-12-05
Add guix.scm file for development of telega on GNU Guix.

brettgilio:
This file makes it easier to contribute to upstream development of telega for people using GNU Guix. It should not conflict with MELPA, and should merge easily. It is a standard practice to use this file to spin up development containers (similar to pyvenv, but far more flexible). Let me know what you think, thanks!

### DIFF
--- a/etc/guix.scm
+++ b/etc/guix.scm
@@ -40,7 +40,7 @@
 	     (srfi srfi-1)
 	     (srfi srfi-26))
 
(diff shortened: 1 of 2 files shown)


brettgilio:
To be more clear. This file is used for development on the latest local checkout of Telega, rather than the version of Telega in upstream Guix.

zevlg:
Merged in, thanks!

PR #116, opened 2019-12-13
Patch `ffplay` and `ffmpeg` paths instead of propagating ffmpeg input.

brettgilio:
This is a revision to the previous pull request. This is a more hygienic way to ensure that ffmpeg functionality is given to emacs-telega on build-time.


zevlg:
Merged in, thanks!

PR #117, opened 2019-12-14
Add GNU Guix badge

brettgilio:
This adds a badge representing the version of Telega available in GNU Guix.

The hyperlink will NOT work yet, because the packages page gets refreshed once every few days, but it will work soon!


zevlg:
Merged in, thanks!

PR #121, opened 2019-12-26
Remove work arounds for imagemagick and svg.

brettgilio:
I have removed this snippet from our local guix.scm recipe. This snippet became unnecessary when switching from building against emacs-minimal to building against emacs in order to provide fringe-bitmap support on bytecompiling.


zevlg:
merged in, thanks!

PR #138, opened 2020-01-15
Add emacs-company as an input to emacs-telega.

brettgilio:
This will provide company-based completion support OOTB for users of Telega on GNU Guix, without having to install company themselves.


brettgilio:
Leaving this open, as per the discussion in the telegram group. This does not modify how it is propagated by MELPA, but is a suggestion for our GNU Guix adjacent package.

PR #149, opened 2020-02-14
[fix] Grammar and consistency in FAQ section of README.md

aidenholmes:


brettgilio:
I think it might be better to replace "I got" with "I have". Wdyt?

aidenholmes:
I agree, should I make the change?

brettgilio:
> I agree, should I make the change?

Yes. In the same branch you made this pull request with, just apply another commit and it will auto-propagate here.

PR #160, opened 2020-04-08
Update `telega-screenshot-function` on macos.

cnsunyour:
On macos:
1) The "screencapture" utility can perform a
complete screenshot operation.
2) The "pngpaste" utility can make image from clipboard.

----

#

### DIFF
--- a/telega-chat.el
+++ b/telega-chat.el
@@ -3103,7 +3103,7 @@ If DOC-P prefix arg as given, then send it as document."
          (temporary-file-directory telega-temp-dir)
          (tmpfile (telega-temp-name "clipboard" ".png"))
          (coding-system-for-write 'binary))
-    (if (eq telega-screenshot-function 'telega-screenshot-with-pngpaste)
+    (if (executable-find "pngpaste")
         ;; NOTE: On MacOS, try extracting clipboard using pngpaste
         (unless (= 0 (telega-screenshot-with-pngpaste tmpfile))
           (error "No image in CLIPBOARD"))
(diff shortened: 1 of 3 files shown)


cnsunyour:
btw,  the "screencapture" utility is included with macos.

telega-util.el
```diff
@@ -872,11 +872,19 @@ If REGION-P is non-nil, then make a screenshot of region."
...
                     nil tofile nil 'quiet))))
 
+(defun telega-screenshot-with-screencapture (tofile &optional _region-p)
+  "Make a screenshot into TOFILE using `screencapture' utility.
+If REGION-P is non-nil, then make a screenshot of region."
+  (let ((pngpaste-cmd (concat (or (executable-find "screencapture")
+                                  (error "Utility `screencapture' not found"))
+                              " -i " tofile)))
+    (call-process-shell-command pngpaste-cmd)))
+
```
zevlg, line 882: Please use "screencapture" (or maybe "screencap") naming here instead of  "pngpaste".

Also I see you always use "-i" flag.  As I remember it means "interactive" i.e. selecting region by hand.  That is exactly the case for use of `region-p`. 
Without `region-p` screenshot function should take screenshot of the whole screen, without interactions with user.
See how `region-p` is used in `telega-screenshot-with-import`

telega-chat.el
```diff
@@ -3125,7 +3125,9 @@ Multiple `C-u' increases delay before taking screenshot of the area."
     (setq n (log (car n) 4)))
 
   (if (and (> n 0)
-           ;; NO delays for "pngpaste"
+           ;; NO delays for "screencapture" or "pngpaste"
+           (not (eq telega-screenshot-function
+                    'telega-screenshot-with-screencapture))
```
zevlg, line 3130: Unlike "pngpaste", "screencapture" supports interactive region selection for the screenshot, so taking delay before taking screenshot with "screencapture" is essential

cnsunyour:
> Thank you for the PR! Please fix the following before merge

Fixed, please check it.

zevlg:
Merged in, thanks!

cnsunyour:
Thanks a lot  for your guidance.

PR #166, opened 2020-05-08
Update README.dm message filtering section

weird-ego:
Small hint to apply filtering to fetched history.


zevlg:
Thanks, however I think this is subject to change.  Defining advices is not good, `telega.el` should provide `hook` way to mark messages as ignored

BTW: we already have `doc/telega-manual.org` for documentation, we should move documentation details into there, keeping `README.md` in some terse style

PR #167, opened 2020-05-09
[enh] better random color generation

BlindingDark:
Current random color algorithm will produce colors that are too dark or uncomfortable, and new algorithm will avoid producing primary colors, too bright or too gray colors.
This is more obvious in dark theme.

Generate color matrix:
```elisp
(defun insert-color (hex)
  (insert (propertize "  " 'font-lock-face `(:background ,hex))))

(defun insert-color-matrix (fun)
  (dotimes (i 10)
    (insert "\n ")
    (dotimes (j 10)
      (insert-color (funcall fun)))
    (insert ";;")))

(insert-color-matrix 'telega-color-random)
```

### DIFF
--- a/telega-util.el
+++ b/telega-util.el
@@ -133,12 +133,26 @@ Selected frame and frame displaying root buffer are examined first."
 (defsubst telega-color-to-hex (col)
   (color-rgb-to-hex (car col) (cadr col) (caddr col) 2))
 
-(defun telega-color-random (&optional lightness)
-  "Generates random color with lightness below LIGHTNESS.
-Default LIGHTNESS is 0.85."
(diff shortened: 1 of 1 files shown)


zevlg:
Thanks for the PR!  But I'm actually moving towards use of `rainbow-identifiers` package for colors generation.  This will be commited soon, I'll let you know after commit

BlindingDark:
It seems that using hash to generate color is a better solution so that the color of same username will remain the same.

If this PR is not useful in the future, please feel free to close it.

BlindingDark:
Since new version using rainbow-identifiers has been released and works well, I will close this PR.

Issue #168, opened 2020-05-12
add ldconfig line in readme.md


nardusg:
## Telega Setup
*OS*: Fedora

## Current Behavior
Some linux distros don't have /usr/local/lib in ldconfig

## Steps to Reproduce
Install telega on Fedora

## Possible Solution
Fixed the readme.md file to include line that ldconfig need to contain /usr/loca/lib

PR #169, opened 2020-05-12
Add readme.md line for ldconfig cache

nardusg:
Hi

This is to add a line to readme.md to warn users to make sure ldconfig is setup correctly.

This pull request is to fix issue:

https://github.com/zevlg/telega.el/issues/168

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,8 @@ $ sudo make install
 It will install headers to `/usr/local/include` and library itself
 into `/usr/local/lib`.  These paths are hardcoded in `telega.el`.
 
+For Linux users, make sure `/usr/local/lib` is in you ldconfig cache, otherwise telega server build will fail.
+
 ### Building libtgvoip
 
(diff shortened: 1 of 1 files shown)


zevlg:
> For Linux users, make sure `/usr/local/lib` is in you ldconfig cache, otherwise telega server build will fail.

Build won’t fail. telega-server will build successfully, however you will get behaviour described in first question in the FAQ. See https://github.com/zevlg/telega.el#faq

nardusg:
Can close it if FAQ is sufficient ?

nardusg:
> > For Linux users, make sure `/usr/local/lib` is in you ldconfig cache, otherwise telega server build will fail.
> 
> Build won’t fail. telega-server will build successfully, however you will get behaviour described in first question in the FAQ. See https://github.com/zevlg/telega.el#faq

It does fail for Fedora users if ldconfig is  not spesified

PR #184, opened 2020-07-05
server/Makefile: Build library rpath into server

dholm:
Builds the rpath for the Telegram library into the `telega-server`
binary so that it will be found automatically when starting
`telega-server`, regardless of where `LIBS_PREFIX` is pointing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,22 +391,6 @@ exceptionally helpful.
 
 # FAQ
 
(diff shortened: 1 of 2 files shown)


zevlg:
Merged in, thanks!  Does this works well on MacOS?

dholm:
It should work the same way on macOS.

PR #187, opened 2020-07-10
Fixup some defcustom variables :type

ojab:
`C-h v telega-use-message-database` -> try to customize -> `Symbol’s function definition is void: nil`. Fix it.


zevlg:
Merged in, thanks @ojab

PR #199, opened 2020-08-24
WIP: New build system

sirikid:
1. Eldev for Elisp part
2. Eldev+Makefile for C/C++ part
3. I tried to write an eldev builder for docs, but nothing decent has worked out so far.


zevlg:
Could you please describe in few words why do we need this?  What problems this PR addresses?

Does these changes to Makefile is BSD make compatible?  Looks like Makefile is GNU make only

sirikid:
> Does these changes to Makefile is BSD make compatible? Looks like Makefile is GNU make only

No. It was not compatible before.

zevlg:
Lets wait a bit TDLib 1.7.0 release, and then come back to the build system.  Probably we need to use `cmake` if we want static linking

Issue #99, opened 2019-10-31
Keep opened only visible chats


zevlg:
When chat gets visibility then call `openChat`, when chat looses visibility, then call `closeChat`

See comments in https://github.com/tdlib/td/issues/758

This can be done using `telega-chatbuf--switch-out` and `telega-chatbuf--switch-in` functions

PR #202, opened 2020-09-03
fix build on non-GNU systems

omar-polo:
First of all, thanks for developing this. I cannot convey in words how happy I am to have a working telegram client on my OpenBSD desktop (telegram web does not count), and it's even within emacs!  You really made my day.

Talking about this proposal: from #199 I can see that some changes to the build system are planned after td 1.7 (I don't know when it's scheduled), so I tried to minimize the changes.  I'm making the path to make customizable, so one can (setq telega-make-command "gmake") before (telega-build-server).  The root Makefile already uses $(MAKE), so this setting is propagated when building the server.

Another possibility could be to change telega-build-server so it let the user change the command to execute when called with a prefix argument.  Now that I'm thinking of it, it's probably better than adding a global variable that can go away soon.

I'm also dropping the -v flag to install(1) as it's non standard and not supported by at least Open and FreeBSD install.

P.S.: apart from this issues with gmake and install, everything else (including photos, notifications, stickers, audio, ...) is working flawlessly!

### DIFF
--- a/server/Makefile
+++ b/server/Makefile
@@ -30,8 +30,8 @@ $(PROG): $(SOURCES)
 	$(CC) $(CFLAGS) -o $(PROG) $(SOURCES) $(LDFLAGS)
 
 install: $(PROG)
-	@install -v -d $(INSTALL_PREFIX)
-	@install -v $(PROG) $(INSTALL_PREFIX)
+	@install -d $(INSTALL_PREFIX)
+	@install $(PROG) $(INSTALL_PREFIX)
 
 test:
 	@python3 run_tests.py
--- a/telega-server.el
+++ b/telega-server.el
@@ -128,7 +128,9 @@ Otherwise query user about building flags."
                )))
       (unless (zerop
                (shell-command
-                (concat "make " build-flags " "
+                (concat (or (executable-find "gmake")
+                            "make")
+                        " " build-flags " "
                         "LIBS_PREFIX=" (expand-file-name telega-server-libs-prefix) " "
                         "INSTALL_PREFIX=" (expand-file-name telega-directory) " "
                         "server-reinstall")))


zevlg:
Thanks, This looks reasonable! But maybe we can modify Makefiles, so they could be compatible with BSD make?

Would it be good solution to use "gmake" if available and fallback to "make" otherwise?

omar-polo:
Zajcev Evgeny <notifications@github.com> writes:

> Thanks, This looks reasonable! But maybe we can modify Makefiles, so they could be compatible with BSD make?

well, I tried to avoid extensive modification to the Makefile after i
read #99.  Most of the makefile is easy to posixify, the only thing I'm
not too sure atm is how to rewrite the WITH_VOID and WITH_TON part.
I'll look at it later today

> Would it be good solution to use "gmake" if available and fallback to "make" otherwise?

This should be even better!  I'm not that proficient in elisp, but would
something like this in telega-server-build be acceptable?

    (concat (or (executable-find "gmake")
                "make")
            " " build-flags " " ...)

zevlg:
> Would it be good solution to use "gmake" if available and fallback to "make" otherwise?
> This should be even better! I'm not that proficient in elisp, but would something like this in telega-server-build be acceptable? (concat (or (executable-find "gmake") "make") " " build-flags " " ...)

exactly! could you please modify it that way and I'll merge then

Thanks!

omar-polo:
Zajcev Evgeny <notifications@github.com> writes:

>> Would it be good solution to use "gmake" if available and fallback to "make" otherwise?
>> This should be even better! I'm not that proficient in elisp, but would something like this in telega-server-build be acceptable? (concat (or (executable-find "gmake") "make") " " build-flags " " ...)
>
> exactly! could you please modify it that way and I'll merge then
>
> Thanks!

I've updated the pr

zevlg:
Merged in, thanks!

PR #213, opened 2020-10-15
add zh.plist, support chinese

toure00:


zevlg:
This is great, I'll take a look how we can integrate this.  In the `telega` distribution we need to translate only values for "telega_XXX" strings, not for "lng_".  Translations for "lng_" strings  are automatically taken from https://translations.telegram.org/en/tdesktop/

We have all the English strings for "lng_" inside etc/langs/en.plist inplace to avoid initial heavy communication with Telegram cloud, this is subject to change in the future.

zevlg:
Merged in, thanks!

PR #215, opened 2020-10-24
key binding 'k' conflict with vim's cursor move up key 'k'

zhigang26:
I am vim user . So I use evil mode to use emacs.

vim move cursor by hjkl.  telega keybinding k conflict with vim.

`telega-msg-delete-marked-or-at-point` bind three keys :  k, d, DEL. so I think,  remove k, there will be no impact on the overall situation.
```
(define-key map (kbd "DEL") 'telega-msg-delete-marked-or-at-point)
(define-key map (kbd "d") 'telega-msg-delete-marked-or-at-point)
(define-key map (kbd "k") 'telega-msg-delete-marked-or-at-point)
```

### DIFF
--- a/telega-msg.el
+++ b/telega-msg.el
@@ -69,7 +69,7 @@
     (define-key map (kbd "e") 'telega-msg-edit)
     (define-key map (kbd "f") 'telega-msg-forward-marked-or-at-point)
     (define-key map (kbd "d") 'telega-msg-delete-marked-or-at-point)
-    (define-key map (kbd "k") 'telega-msg-delete-marked-or-at-point)
+    ;; (define-key map (kbd "k") 'telega-msg-delete-marked-or-at-point)
     (define-key map (kbd "l") 'telega-msg-redisplay)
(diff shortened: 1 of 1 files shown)


cireu:
This is a keybinding related to `evil`, There're also lots of people use Emacs without evil, this may break their habit.

Can you please use 

```emacs-lisp
(with-eval-after-load 'telega
  (define-key telega-msg-button-map "k" nil))
```
to do your customization.

You can also contribute a `evil-telega.el` overlay to https://github.com/emacs-evil/evil-collection to help other evil fans.

zevlg:
This is out of the telega.el scope, if you need <kbd>k</kbd> key to do something else in your custom setup, you need to do it in your setup, not in telega.el code. For example:
```
(define-key telega-msg-button-map (kbd "k") nil)
```

There are too many custom setups to handle them in telega.el code

zhigang26:
After communicating with some emacs users, the concept of k(ill) is important to them. My PR only considers vimer's feelings. I accept your suggestions, I appreciate your solution.

PR #220, opened 2020-10-30
Use telega-username as the face of tracking mode-line notifications

jaor:
When adding a buffer for notifications with the tracking package, one can specify a list of faces to use for it in the modeline.  This is useful not only for appearance, but also because the face can be used to prioritize the notification, altering the order in which it is shown (via tracking-faces-priorities).  I think telega-username is a good candidate for this, at least as a default.

### DIFF
--- a/telega-customize.el
+++ b/telega-customize.el
@@ -144,7 +144,9 @@ where PROXY-TYPE is one of:
 (defcustom telega-use-tracking-for nil
   "*Specifies Chat Filter for chats to be tracked with tracking.el.
 Make sure you have tracking.el loaded if this option is used.
-Only chats with corresponding opened chatbuf are tracked."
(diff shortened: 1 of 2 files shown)


zevlg:
This is nice!  However I think this should be customizable, i.e. we should have dedicated `telega-tracking` face for this purpose.  `telega-username` face might look ok in chatbuf, but not ok in modeline.

Could you please add `telega-tracking` face?  Also, could you please mention this face in docstring to `telega-use-tracking-for`, so I could regenerate telega manual mentioning this functionality.

Thanks

jaor:
i've amended the patch with the suggested modifications (we can change
the default for telega-tracking if you feel telega-username is not
adequate, just let me know).

thanks!

zevlg:
Merged in, thanks @jaor !

PR #225, opened 2020-11-24
fix nativecomp emacs cannot be used normally

Makima0:
nativecomp produces the following error
`
error in process filter: telega-emoji-create-svg: Invalid function: telega-svg-create
error in process filter: Invalid function: telega-svg-create
Error running timer ‘telega-server--idle-timer-function’: (invalid-function telega-ins-fmt)
`


zevlg:
Thanks, merged in!

PR #230, opened 2020-12-16
Use dbus-ignore-errors in telega-notifications--close.

cireu:
Now can set `dbus-debug` to `t` if somebody want to debug with notification
related issues.


telega-notifications.el
```diff
@@ -31,6 +31,7 @@
 
 ;;; Code:
 (require 'cl-lib)
+(require 'dbus)
```
zevlg, line 34: Is not this explicitly required in `notifications' ?  Should we really need this?

cireu: Yes, confirm that we don't need this. removed.

zevlg:
Merged in, thanks!

PR #221, opened 2020-11-01
Use version< instead of string< when comparing version.

BlindingDark:
```
Warning (emacs): TDLib version=1.6.10 < 1.6.9 (min required), please upgrade TDLib and recompile ‘telega-server’
```

But it seems to me that version 1.6.10 is not compatible.

```
[...]

PR #234, opened 2020-12-24
Small addition to docs

digital-scrappy:
I added a small section to the docs that covers a problem I ran into. 
Basically I did not realize that re-installing telega via Melpa  would not rebuild the server so I received the minimum TDLib version warning even though the correct version was installed.

### DIFF
--- a/docs/telega-ellit.org
+++ b/docs/telega-ellit.org
@@ -828,6 +828,4 @@ project.
 * Troubleshooting
 ** Warning for TDLib version below minimum required, even though correct version is installed
 If at starting telega you get the warning =TDLib version=%s < %s (min required),please upgrade TDLib and recompile `telega-server= even though your tdlib version is satisfying telega's requirements, be aware that deleting the telega package and reinstalling it via melpa will not recompile the server, run =telega-server-build= instead.
-
 #  LocalWords:  customizable chatbuf tdlib gmake
-
--- a/docs/telega-ellit.org
+++ b/docs/telega-ellit.org
@@ -825,4 +825,9 @@ project.
 
 #+ELLIT-INCLUDE: ../contrib/telega-mnz.el
 
+* Troubleshooting
+** Warning for TDLib version below minimum required, even though correct version is installed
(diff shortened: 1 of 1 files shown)


zevlg:
Maybe you run into the problem fixed in https://github.com/zevlg/telega.el/pull/221 ?

digital-scrappy:
I don't think so. I originally had a quite old version of TDLib and the warning persisted after I had rebuilt TDLib from their git and had reinstalled telega using Melpa. Could it be that the telega-server does not get rebuilt when telega is reinstalled?.

zevlg:
`telega-server` not necessary need to be rebuild on `telega` updates, only if `telega-server-min-version` is greater then version of the `telega-server` currently installed.  Check your telega-server version with <kbd>M-: (telega-server-version) RET</kbd> and value for `telega-server-min-version` with <kbd>C-h v telega-server-min-version RET</kbd>

After updating TDLib, do not forget reloading libs cache with `ldconfig`, then check telega-server version with `$ telega-server -h`

zevlg:
> I don't think so. I originally had a quite old version of TDLib and the warning persisted after I had rebuilt TDLib from their git and had reinstalled telega using Melpa. Could it be that the telega-server does not get rebuilt when telega is reinstalled?.
> Just to be sure I don't think this is a problem since the warning specifies that one should rebuild the server. I just thought it might be confusing to others (as it was to me) that the rebuild does not happen upon reinstall.

telega should query you about `telega-server` rebuild if rebuild is needed, that is done by `(telega-server--check-version telega-server-min-version)`

If it does not, then there is some bug that should be fixed.

Could you check please, also check your telega version, because there was issue related to this, fixed in https://github.com/zevlg/telega.el/pull/221

digital-scrappy:
hi my telega version is 1.7.0 and my telega-server version is 0.7.3

zevlg:
> hi my telega version is 1.7.0 and my telega-server version is 0.7.3

That is ok, you can run telega with such versions.

However, I’ve been asking for version values when you had problems and telega does not query you for server rebuild as you described in first message

digital-scrappy:
Ah ok, sorry for not being specific, I don't know for sure what my tdlib version was (tried searching for it in my logs but did not find it). I guess the  telega version must have been the same since I had just installed it that day (the 24th of December).

PR #247, opened 2021-01-11
Rename the function telega-chatbuf-attach-member to something more intuitive.

yur3i:
While "attach-member" might be consistent with the other "attach-*" functions, given the nature of the command and the consequences of misuse as well as the lack of confirmation it is important to make it clear that this adds a NEW member rather than mentioning someone who is already a member.

### DIFF
--- a/telega-chat.el
+++ b/telega-chat.el
@@ -3403,7 +3403,7 @@ Uses `telega-screenshot-function' to take a screenshot."
         (x-focus-frame (window-frame (get-buffer-window)))
         (telega-chatbuf--attach-tmp-photo tmpfile)))))
 
-(defun telega-chatbuf-attach-member (user)
+(defun telega-chatbuf-add-member (user)
   "Add USER to the chat members."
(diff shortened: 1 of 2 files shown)


zevlg:
Probably better to remove this attachment type at all and use `M-x telega-chat-add-member RET` to add chat members?

yur3i:
I did think that but if there was a reason that it was in there that I overlooked then I didn't want to mess with it

telega-chat.el
```diff
@@ -3403,7 +3403,7 @@ Uses `telega-screenshot-function' to take a screenshot."
         (x-focus-frame (window-frame (get-buffer-window)))
         (telega-chatbuf--attach-tmp-photo tmpfile)))))
 
-(defun telega-chatbuf-attach-member (user)
+(defun telega-chatbuf-add-member (user)
```
zevlg, line 3406: This function not needed anymore, delete it as well.  There is `telega-chat-add-member` command exists

zevlg:
merged in, thanks!

PR #258, opened 2021-01-29
Make native-comp happy

yangsheng6810:
When using native-comp branch of Emacs, elisp codes are compiled to native code. Some macros used in telega were not correctly required at compile time, causing errors like the following for some users:
```
---[telega bug]
PP-ERROR: (telega-root--chat-known-pp OMITTED_FOR_PRIVACY
) ==>
(invalid-function dom-attr)
------
```

### DIFF
--- a/telega-modes.el
+++ b/telega-modes.el
@@ -29,6 +29,7 @@
 (require 'telega-server)
 (require 'telega-filter)
 (require 'telega-util)
+(eval-when-compile (require 'dom))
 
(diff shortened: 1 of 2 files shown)


zevlg:
Is not https://github.com/zevlg/telega.el/pull/256 fixed this ?

yangsheng6810:
Sorry, it does fix it. I forgot to update telega before submitting the PR.

PR #284, opened 2021-04-21
Use canonical test for within a minibuffer

yangsheng6810:
The canonical way to test wihin a minibuffer is via `minibufferp`. The
old way of `(eq major-mode 'minibuffer-inactive-mode)` will no longer
work after 55db25b257 in Emacs master branch (see Emacs bug 47150 for
details). This commit fixes an old bug related to the implementation of
when is the minibuffer in minibuffer-inactive-mode. After the commit, we
have minibuffer-mode and minibuffer-inactive-mode.


zevlg:
merged in, thanks!

PR #295, opened 2021-06-15
Add msg saved dir for saving without asking

Vonfry:


telega-customize.el
```diff
@@ -1381,6 +1381,12 @@ no longer work."
   :type 'boolean
   :group 'telega-msg)
 
+(defcustom telega-msg-saved-dir nil
```
zevlg, line 1384: Better to name it `telega-msg-save-dir` or even better `telega-msg-save-directory`

Also, first line of the doc string must end with ".". Put things starting with "If nil .." on next line.

Vonfry: Change to `telega-msg-save-dir` which has similar format as `telega-temp-dir`.

zevlg:
Merged in, thanks for your contribution!

PR #299, opened 2021-07-09
Do not attempt making any images when telega-use-images is nil

akater:
TUI support is much better than it used to be.  However, `<IMAGE>` placeholder is never to be seen as image-creating procedures still run, and chat buffer gets filled with `PP-ERROR` output.

The patch prevents them from running.

Testing with 0.7.025, works fine so far.

### DIFF
--- a/telega-ins.el
+++ b/telega-ins.el
@@ -543,12 +543,14 @@ SHOW-DETAILS - non-nil to show photo details."
             (telega-ins (propertize "Self-descruct in" 'face 'shadow) " "
                         (telega-duration-human-readable ttl-in) "\n"))
           (telega-ins--image-slices
-           (telega-self-destruct-create-svg
-            (plist-get photo :minithumbnail)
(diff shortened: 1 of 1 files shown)


zevlg:
This patch might do the job, and telega already have some similiar
code workarounding issue with emacs-nox, however, this is not quite a
correct way to utilize `telega-use-images`.  Images oriented code
should be agnostic about `telega-use-images` and all things that
depend on `telega-use-images` should be handled in other places, such
as `telega-media--create-image` or/and `telega-ins--image-slices`.

Set `debug-on-error` to non-nil to get the backtrace instead of
`PP-ERROR` and please report an issue providing your setup info,
telega version and the backtrace you get

Sorry for not accepting your PR once again, however, `telega` code is already pretty complex to maintain, adding conditions to the code will complicate it more.  We need to remove conditional code, not adding it

akater:
It's OK, I did suspect it won't fly but I prefer providing a workaround, even if hackish, whenever I can.

akater:
The problem is gone now.

PR #322, opened 2021-11-01
[enh] Support dired-dwim-target

akater:
Optionally guess default directory from dired buffers when picking
attachments.  By default, inherit user's `dired-dwim-target`
setting (dynamically).

In `telega-describe-chat--inserter` you set dir to nil explicitly.  I did not change that.

### DIFF
--- a/telega-chat.el
+++ b/telega-chat.el
@@ -654,17 +654,6 @@ Specify non-nil BAN to ban this user in this CHAT."
   (interactive (list telega-chatbuf--chat))
   (telega--unpinAllChatMessages chat))
 
-(defun telega--read-file-name (prompt &optional dir default-filename
-                                      mustmatch initial predicate)
-  "Like `read-file-name' but taking `telega-dired-dwim-target' into account."
-  (read-file-name prompt
-                  (cl-ecase telega-dired-dwim-target
-                    (inherit (dired-dwim-target-directory))
-                    (t (let ((dired-dwim-target t))
-                         (dired-dwim-target-directory)))
-                    ((nil) nil))
-                  default-filename mustmatch initial predicate))
-
 (defun telega-describe-chat--inserter (chat)
   "Inserter for the CHAT description."
   (let ((chat-ava (telega-msg-sender-avatar-image chat)))
@@ -694,7 +683,7 @@ Specify non-nil BAN to ban this user in this CHAT."
       (telega-ins " ")
       (telega-ins--button "Set Profile Photo"
         'action (lambda (_ignored)
-                  (let ((photo (telega--read-file-name "Profile Photo: " nil nil t)))
+                  (let ((photo (telega-read-file-name "Profile Photo: " nil nil t)))
                     (telega--setProfilePhoto photo)))))
     (when (telega-chat-match-p chat '(my-permission :can_invite_users))
       (telega-ins " ")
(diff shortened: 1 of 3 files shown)


telega-chat.el
```diff
@@ -654,6 +654,17 @@ Specify non-nil BAN to ban this user in this CHAT."
   (interactive (list telega-chatbuf--chat))
   (telega--unpinAllChatMessages chat))
 
+(defun telega--read-file-name (prompt &optional dir default-filename
```
zevlg, line 657: This should be in the `telega-util.el` file along with other `telega-read-XXX` functions.  Make it public, i.e. `telega-read-file-name`

telega-chat.el
```diff
@@ -654,6 +654,17 @@ Specify non-nil BAN to ban this user in this CHAT."
...
+                                      mustmatch initial predicate)
+  "Like `read-file-name' but taking `telega-dired-dwim-target' into account."
```
zevlg, line 659: Value for the `dir` argument should be regarded, i.e. second argument for the `read-file-name` should be in the form `(or dir XXX)`, so if `dir` is explicitly specified it should take precedence before other logic

telega-customize.el
```diff
@@ -1192,6 +1192,15 @@ COMMAND-FUNC - Command function to execute."
   :type 'list
   :group 'telega-chat)
 
+(defcustom telega-dired-dwim-target 'inherit
```
zevlg, line 1195: Sounds a bit complicated. Maybe make its non-nil value to be value for the `dired-dwim-targed` in the `telega-read-file-name` ?   This will simplifies code overall, like:

```
(defcustom telega-dired-dwim-target dired-dwim-target ...)
```
and then in the `telega-read-file-name` we will have
```
(or dir
    (when telega-dired-dwim-target
      (let ((dired-dwim-target telega-dired-dwim-target))
        (dired-dwim-target-directory))))
```
for the second argument to the `read-file-name`

zevlg (changes requested):

telega-customize.el
zevlg: Or maybe create `telega-read-file-dir-function` custom variable instead to be a function returning directory?  So it could be `dired-dwim-target-directory` if user wants to utilize `dired-dwim-target` option

This looks like most flexible way, so someone could implement other logic, for example looking at file read history and taking most usable base directory, or any other logic

akater: This option is anticipated in `dired-dwim-target` proper and some reasonable choices are offered in its `custom` interface so it's arguably preferrable to reuse it via `custom-variable-type` which would also automatically adjust to dired-dwim feature as provided by Emacs, whatever the Emacs version is.

If `custom-declare-variable` is appropriate, this amounts to

```lisp
(custom-declare-variable
 'telega-dired-dwim-target
 dired-dwim-target
 "*Value to bind `dired-dwim-target' to, in telega file pickers."
 :type
 (custom-variable-type 'dired-dwim-target)
 :group 'telega-chat)
```

Since `custom-declare-variable` is public, I consider this the best approach.

zevlg: Okay, I'm good with either way: `telega-dired-dwim-target` with `custom-descrare-variable` or just `telega-read-file-dir-function` user option.

But keeping `(when telega-dired-dwim-target ...)` is essential I think, because we do not rely on `dired-dwim-target-directory` logic returning nil, just because major mode is not dried.  We rather explicitly specify that if `telega-dired-dwim-target` is unset, then `dir` is nil

zevlg:

PR #324, opened 2021-11-08
[fix] `custom-declare-variable' wants a form rather than a value

akater:
It's even highlighted in the docstring of `custom-declare-variable` but I missed it:

> DEFAULT should be an expression to evaluate to compute the default value,
> not the default value itself.


zevlg:
Merged in, thanks!

PR #333, opened 2021-12-14
[add] Dynamic check of supported audio inputs

IaH0nii3Re:
telega-vvnote-voice-record-args, telega-vvnote-video-record-args
telega-msg-open-video-note use introdused telega-vvnote--has-audio-inputs
to dynamicaly check the supported audio inputs.

### DIFF
--- a/telega-ffplay.el
+++ b/telega-ffplay.el
@@ -27,22 +27,35 @@
 (require 'cl-lib)
 (require 'telega-core)
 
+(defun telega-ffplay--parse-ffmpeg-output (option support &rest desired-properties)
+  "Parsing the ffmpeg output for the OPTION for the DESIRED-PROPERTIES.
+SUPPORT should be a list of states supported by the DESIRED-PROPERTIES.
+Currently it may contain a decoder, an encoder, or both."
+  (let* ((output (shell-command-to-string (concat "ffmpeg -v quiet -" option)))
+         (header-end (string-match-p "-" output))
+         (additional-characters
+          (- (string-match-p
+              "="
+              (string-trim
+               (cadr (split-string (substring output 0 header-end) "\n"))))
+             2)))
+    (cl-remove-if-not
+     #'stringp
+     (mapcar (lambda (prop)
+               (when (string-match-p
+                      (concat (concat (if (memq 'decoder support) "D" ".")
+                                      (if (memq 'encoder support) "E" "."))
+                              (when (> additional-characters 0)
+                                (make-string additional-characters ?\.))
+                              prop)
+                      output)
+                 prop))
+             desired-properties))))
+
 (defun telega-ffplay-check-codecs (how &rest codecs)
   "Check CODEC is available in ffmpeg.
 Return list of available codecs."
-  (let ((codecs-output
-         (shell-command-to-string "ffmpeg -v quiet -codecs")))
-    (cl-remove-if-not
-     #'stringp
-     (mapcar (lambda (codec)
-               (when (string-match-p
-                      (concat (concat (if (memq 'decoder how) "D" ".")
-                                      (if (memq 'encoder how) "E" "."))
-                              "....."
-                              codec)
-                      codecs-output)
-                 codec))
-             codecs))))
+  (apply #'telega-ffplay--parse-ffmpeg-output "codecs" how codecs))
 
 (defconst telega-ffplay--has-encoders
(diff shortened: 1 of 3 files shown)


telega-ffplay.el
```diff
@@ -27,22 +27,35 @@
...
-                      (concat (concat (if (memq 'decoder how) "D" ".")
-                                      (if (memq 'encoder how) "E" "."))
-                              "....."
-                              codec)
-                      codecs-output)
-                 codec))
-             codecs))))
+                      (concat (concat (if (memq 'decoder support) "D" ".")
+                                      (if (memq 'encoder support) "E" "."))
+                              (when (> additional-characters 0)
```
zevlg, line 48: Maybe remove this `when` condition, because `(make-string 0 ?\.)` returns empty string.  And add assertion that `additional-characters` must be `>=0` ?

telega-ffplay.el
```diff
@@ -27,22 +27,35 @@
...
-(defun telega-ffplay-check-codecs (how &rest codecs)
-  "Check CODEC is available in ffmpeg.
-Return list of available codecs."
-  (let ((codecs-output
-         (shell-command-to-string "ffmpeg -v quiet -codecs")))
+(defun telega-ffplay--parse-ffmpeg-output (option support &rest desired-properties)
+  "Parsing the ffmpeg output for the OPTION for the DESIRED-PROPERTIES.
+SUPPORT should be a list of states supported by the DESIRED-PROPERTIES.
+Currently it may contain a decoder, an encoder, or both."
+  (let* ((output (shell-command-to-string (concat "ffmpeg -v quiet -" option)))
```
zevlg, line 34: Left this "-" to the option, so you will pass "-codecs" and "-devices" to this func, instead of "codecs" and "devices" which looks a bit confusing.

telega-ffplay.el
```diff
@@ -27,22 +27,35 @@
 (require 'cl-lib)
 (require 'telega-core)
 
-(defun telega-ffplay-check-codecs (how &rest codecs)
-  "Check CODEC is available in ffmpeg.
-Return list of available codecs."
-  (let ((codecs-output
-         (shell-command-to-string "ffmpeg -v quiet -codecs")))
+(defun telega-ffplay--parse-ffmpeg-output (option support &rest desired-properties)
```
zevlg, line 30: "parse" is not a good choice for the function name.  "parse" means some *transformation* of something from unstructured representation to structured.  In our case, we do parsing and then checking the structure, so I think "check" or "parse-and-check" is a much better name than simple "parse".

telega-vvnote.el
```diff
@@ -51,11 +51,20 @@
   :type 'boolean
   :group 'telega-vvnote)
 
+(defun telega-vvnote-check-devices (how &rest devices)
+  "Check DEVICES is available in ffmpeg.
+Return list of available DEVICES."
+  (apply #'telega-ffplay--parse-ffmpeg-output "devices" how devices))
```
zevlg, line 57: Use "-devices" here instead of "devices", see comment above about "-" prefix

telega-vvnote.el
```diff
@@ -73,7 +82,8 @@
   (concat (if (eq system-type 'darwin)
               "-f avfoundation -s 640x480 -framerate 30 -i default -r 30 -f avfoundation -i :default"
             ;; gnu/linux
-            "-f v4l2 -s 320x240 -i /dev/video0 -r 30 -f alsa -i default")
+            (concat "-f v4l2 -s 320x240 -i /dev/video0 -r 30 -f "
```
zevlg, line 85: Put `"-f " (car telega-vvnote--has-audio-inputs) " -i default"` on a separate line, i.e.
```
(concat "-f v4l2 -s 320x240 -i /dev/video0 -r 30 "
        "-f " (car telega-vvnote--has-audio-inputs) " -i default")
```

zevlg (changes requested):
Thanks for the contribution!  We need to polish minor things before merging in, see the review comments

telega-ffplay.el
IaH0nii3Re: Yes, that's better.

telega-ffplay.el
IaH0nii3Re: I agree.

telega-ffplay.el
IaH0nii3Re: Yes, "parse" isn't the best choice here, I'll replace it with a "check"

zevlg:

PR #344, opened 2022-01-17
[fix] Use a function with one argument for the "Change" button action.

IaH0nii3Re:
Hi, in \*Telega chat info\* buffer, when I click the "Change" button, I get this error: telega-button--action: Wrong number of arguments: (2 . 2), 1. This pull-request fixes that.

### DIFF
--- a/telega-chat.el
+++ b/telega-chat.el
@@ -607,18 +607,22 @@ Specify non-nil BAN to ban this user in this CHAT."
      (list chat (read-string "New title: " (telega-chat-title chat)))))
   (telega--setChatTitle chat title))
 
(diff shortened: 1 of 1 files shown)


zevlg:
Nice catch!  Do you mind I'll fix this in a bit different manner?  Also, Telegram supports 1month as TTL, we need to add support for this as well

`telega-chat-set-ttl` is just an interactive version of `telega--setChatMessageTtl` function, so we can always read ttl from user, removing `ttl` argument

IaH0nii3Re:
> Nice catch! Do you mind I'll fix this in a bit different manner? 
Of course, you can fix it better.

PR #395, opened 2023-03-03
Introduce contrib/telega-bridge-bot.el

BlindingDark:
Hi there.

This package will replace the message bridge bot (like @matrix_t2bot) to make it look like the user itself.
Please have a look!

Thanks! :pray:

### DIFF
--- a/contrib/telega-bridge-bot.el
+++ b/contrib/telega-bridge-bot.el
@@ -1,0 +1,516 @@
+;;; telega-bridge-bot.el --- Replace bridge bot user        -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2023 by Henry Sun.
+
+;; Keywords:
+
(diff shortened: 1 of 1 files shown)


zevlg (changes requested):
Thank you for this awesome contribution!  I've added few comments on the code.

In general, it is better to use objects itself, such as `chat`, `user`, `message` instead of using id to reference these objects.  Properties of the objects changes sometimes in the TDLib, so code that references object properties directly is much more fragile in comparison to code that uses abstractions provided by `telega`.

It is fully okay to have this code as initial commit, however, I think the default value for `telega-bridge-bot-bridge-info-plist` user option should not be forced and should be configured by user of the package

BlindingDark:
Hi @zevlg . Thanks for your advice, I just pushed the changes, please have a look.

zevlg:
Merged in, thank you for this contribution!

PR #396, opened 2023-03-04
[doc] Add contrib/telega-bridge-bot.el

BlindingDark:
Add document of the `contrib/telega-bridge-bot.el` to `telega-ellit.org`


zevlg:
Thanks, documentation will be regenerated on my next commit in few days

PR #407, opened 2023-03-20
upgrade qemu and buildx to support arm64

grburgess:
these actions should build arm and amd archs, I bounced them to the latest version (v2)... in principle that should all work.


zevlg:
Merged in, thanks for the contribution!  I'll rebuild images in few minutes using this updated pipeline

grburgess:
no worries. Not sure if it fixes it, but if not I think I know another option

grburgess:
seems that didn't work. I'll try some more options

Issue #317, opened 2021-09-23
Images with stripes


vitek:
## Telega Setup
*OS*: Ubuntu 20.04.2 LTS
*Emacs*: GNU Emacs 28.0.50 (build 3, x86_64-pc-linux-gnu, GTK+ Version 3.24.20, cairo version 1.16.0)
*Telega*: telega v0.7.70 (TDLib v1.7.8) (telega-server v0.7.13)
*MELPA*: 20210922.739

## Current Behavior

Sometimes images are split by text lines with ugly gaps between them. That often happens with user pic: 

![image](https://user-images.githubusercontent.com/489439/134457326-21f1c158-dd2c-4565-865d-19ced7720fa3.png)

Things get worse when `text-scale` is increased in runtime:

![image](https://user-images.githubusercontent.com/489439/134457367-aad055a6-1422-49e3-a646-bcdc00dba51d.png)

zevlg:
As to avatar, this is most possibly due to "⮪" (`telega-symbol-reply`) character, that causes line height to grow.

If you have emoji font installed, so your automatic value for `telega-emoji-use-images` is non-nil, then you can add `reply` to the `telega-symbols-emojify` list, so you will get image instead of direct character.  Or you can set a different value for `telega-symbol-reply` that won't grow your line height.

As to `text-scale`, this is not supported in runtime, however if you set buffer's default face in the `telega-chat-mode-hook`, then  everything should work.  See https://github.com/zevlg/telega.el/issues/90

vitek:
> As to avatar, this is most possibly due to "⮪" (telega-symbol-reply) character, that causes line height to grow.

That helped. But there are other font issues, e.g.:

![image](https://user-images.githubusercontent.com/489439/135155105-3089ff69-14f6-46f4-a639-2af51aad4f18.png)

Is it possible to draw it as whole images without splitting into lines?

> As to text-scale, this is not supported in runtime, however if you set buffer's default face in the telega-chat-mode-hook, then everything should work. See #90

That's sad. Is it possible to install hook that will adjust telega's internal state on zoom in/out?

zevlg:
> > As to avatar, this is most possibly due to "⮪" (telega-symbol-reply) character, that causes line height to grow.
> 
> That helped. But there are other font issues, e.g.:
> 
> ![image](https://user-images.githubusercontent.com/489439/135155105-3089ff69-14f6-46f4-a639-2af51aad4f18.png)
> 
[...]

PR #410, opened 2023-03-22
Adjust the ascent to avoid gaps between avatar slices.

BlindingDark:
This PR tries to calculate the maximum decent for the characters used in the username, and then use that to set the ascent for the upper half slice of the avatar.
For the slice of the lower half of the avatar, it is unrealistic to calculate the ascent of the line where it is, because for example, the content of the reply message has not been fetched when the avatar is inserted. So I decided to provide an option `telega-avatar-slice-2-ascent`, the default is 100 which is the maximum value of ascent, which is to do our best to make it higher than other text. And even setting it to 100 seems to have almost no visual impact. Anyway, no harm in having it as a custom option.

This PR can solve most of the avatar gaps problems, such as this one mentioned in the following issue.
image
https://github.com/zevlg/telega.el/issues/317#issuecomment-930042574

For the avatar of the three lines in `@replies`, I choose to not touch the original logic, that is keep the ascent as `'center`, because we did not calculate the ascent of the second line, so we cannot get the correct height for the third line.

Please let me know if you have any suggestions, thanks! :pray:


BlindingDark:
Currently using this implementation will not be able to update the avatar, which may affect existing functionality, so I decided to close this PR.

PR #413, opened 2023-04-18
Add support for libayatana-appindicator library to telega-server

jumper047:
Ubuntu 22 drops support of the `libappindicator` library in prior to `libayatana-appindicator`. It still has `libappindicator` in it's repo, but `ubuntu-desktop` package depends on ayatana version, so user can't just install `libappindicator`. You can check some background here - https://github.com/AyatanaIndicators/libayatana-appindicator/issues/30#issuecomment-950224779. 

Luckily, `libayatana-appindicator` interface is identical to `libappindicator`, so I added some additional ifs to use any of those libraries with `telega-appindicator`.

### DIFF
--- a/server/Makefile
+++ b/server/Makefile
@@ -18,6 +18,12 @@ CFLAGS+=-DWITH_APPINDICATOR
 CFLAGS+=`pkg-config --cflags appindicator3-0.1`
 LDFLAGS+=`pkg-config --libs appindicator3-0.1`
 SOURCES+=telega-appindicator.c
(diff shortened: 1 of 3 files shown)


jumper047:
> revert changes to doc/

done

> does etc/Dockerfile also requires that new appindicator library?

I'll try to test it - if I understand it right, telega-server manipulates tray icon via  `appindicator` library inside container. If this statement is correct , then, maybe, it's worth to update it because new library is maintained by community and old one - not.

zevlg:
Merged in, thank you for the contribution!

PR #418, opened 2023-06-18
[telega-bridge-bot] fix: get text without format

BlindingDark:
Username may not be a plain string, for example

```emacs-lisp
#("     " 0 2 (telega-display "𝐿") 2 4 (telega-display "𝑢") 4 6 (telega-display "𝑐") 6 8 (telega-display "𝑖") 8 10 (telega-display "𝑢") 10 12 (telega-display "𝑠"))
```

This can cause problems since we use it as a hash table key.
So here we use `(telega-tl-str content-text :text 'no-props)` to get the string without props.


zevlg:
Merged in, thanks!

PR #419, opened 2023-06-20
[telega-bridge-bot] better way to handle formatted text.

BlindingDark:
This PR replaces #418 with a better way of handling fmt.

In this PR, we are not converting the whole message to no props format, but only when using the username as the hash key.
This avoids losing formatting information in the body.


zevlg:
merged in, thanks!

PR #422, opened 2023-08-10
Two ~ are required in markdown2 mode for strikethrough

samueltardieu:


samueltardieu:
I've only modified the source file, I don't have the environment setup to regenerate other files.

zevlg:
Merged in, thanks! I'll regenerate the docs

PR #425, opened 2023-08-19
[telega-bridge-bot] fix: Update telega-user-title function usage.

BlindingDark:
The `telega-user-title` function has modified in recent updates, so we also need update here.

Here we display both full_name and bot username like "Forward: Hello alice@matrix_bot",


zevlg:
merged in, thanks!

PR #426, opened 2023-08-19
Remove duplicate spaces when user has only first_name.

BlindingDark:
If the user only sets the first_name, there will be extra spaces displayed.
This PR intends to fix it.

### DIFF
--- a/telega-user.el
+++ b/telega-user.el
@@ -132,7 +132,7 @@ Return nil if given FMT-TYPE is not available."
                       (let ((first-name (telega-tl-str user :first_name))
                             (last-name (telega-tl-str user :last_name)))
                         (concat first-name
-                                (when first-name
+                                (when (and first-name last-name)
                                   " ")
                                 last-name)))
                      ((eq fmt-type 'username)


telega-user.el
```diff
@@ -131,10 +131,8 @@ Return nil if given FMT-TYPE is not available."
                      ((eq fmt-type 'full-name)
                       (let ((first-name (telega-tl-str user :first_name))
                             (last-name (telega-tl-str user :last_name)))
-                        (concat first-name
-                                (when first-name
-                                  " ")
-                                last-name)))
+                        (mapconcat #'identity
```
zevlg, line 134: This version generates more garbage, probably something like `(when (and first-name last-name)` as condition for extra space is better?

zevlg (changes requested):

telega-user.el
BlindingDark: Thanks @zevlg, I just updated the code, please review :pray:

zevlg:
Merged in, thanks for your effort!

PR #431, opened 2023-09-05
Update telega-manual.org

dertuxmalwieder:
Fixed TDLib installation with Homebrew. By default, 1.8.0 is installed *unless* you pass `--HEAD`.


zevlg:
`telega-manual.org` is automatically generated file, do not edit it by hand.  I'll add the change to the manual, thank you!

PR #432, opened 2023-10-03
telega-ins: be resilient to missing story/story sender not defined when inserting a webpage

seblemaguer:
Hello,

I faced an error when receiving a link as `story_id` and `story_sender_chat_id` are both nil. I don't know if it is due to the use of emacs 29[1], but by simply testing if the value is available in addition to different to 0, it seems to solve the problem

[1] GNU Emacs 29.1.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.38, cairo version 1.17.8) of 2023-09-01

### DIFF
--- a/telega-ins.el
+++ b/telega-ins.el
@@ -1021,8 +1021,8 @@ Return `non-nil' if WEB-PAGE has been inserted."
               ((plist-get web-page :voice_note)
                (telega-ins--voice-note msg (plist-get web-page :voice_note))
                (telega-ins "\n"))
-              ((and (not (zerop (plist-get web-page :story_sender_chat_id)))
-                    (not (zerop (plist-get web-page :story_id))))
+              ((and (not (telega-zerop (plist-get web-page :story_sender_chat_id)))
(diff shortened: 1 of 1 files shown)


telega-ins.el
```diff
@@ -1021,7 +1021,9 @@ Return `non-nil' if WEB-PAGE has been inserted."
               ((plist-get web-page :voice_note)
                (telega-ins--voice-note msg (plist-get web-page :voice_note))
                (telega-ins "\n"))
-              ((and (not (zerop (plist-get web-page :story_sender_chat_id)))
+              ((and (plist-get web-page :story_sender_chat_id)
```
zevlg, line 1024: To fix this, just use `telega-zerop` instead of `zerop` it will treat nil values as zero

zevlg (changes requested):

seblemaguer:
Thanks, I didn't know this function :D 

I replaced the commit to use `telega-zerop` instead of `zerop`

PR #439, opened 2023-10-31
[contrib/bridge-bot] Fix: telega-msg--pp is now telega-chatbuf-msg--pp

BlindingDark:
`telega-msg--pp` has been renamed to  `telega-chatbuf-msg--pp` in recent updates.


zevlg:
Merged in, thanks!

PR #443, opened 2023-11-11
[WIP] Support tdlib v1.8.21

BlindingDark:
Hi there. This is my personal attempt to support the new tdlib. Please feel free to leave some comments and use/modify the code here.

### TODOs
- [X] Support display quote reply.
- [ ] Highlight the quoted text when jump to the replied msg.
- [ ] Display info (maybe in the echo area) if it can't find the quoted text when jump to the replied msg.
- [ ] Quote only a part of text when reply.
- [ ] Support code language
- [X] Rename some names to support new tdlib


zevlg:
Thanks for your effort!  Unfortunately support for TDLib 1.8.21 is much more complex, than provided in this PR, even for initial support.

Be patient, you will get support for TDLib 1.8.21 soon

zevlg:
I've pushed initial support for TDLib 1.8.21, please check it out

PR #449, opened 2023-11-17
[contrib/telega-bridge-bot] Support tdlib 1.8.21

BlindingDark:
Changes
- Now the forward chat id should be get in `:forward_info` `:from_chat_id`
- Display `fullname @username` in the reply message when that user is a bridge bot. Because the username of bridge bot are the same, so fullname needs to be added in order to distinguish them.


zevlg:
Merged in, thanks!

PR #450, opened 2023-11-17
Highlight quote when jumping to message

BlindingDark:
This will behave more like the official Telegram client.

### DIFF
--- a/telega-chat.el
+++ b/telega-chat.el
@@ -5768,18 +5768,33 @@ Return non-nil on success."
           (pulse-momentary-highlight-region
            (button-start msg-button) (button-end msg-button))
           (when (stringp highlight)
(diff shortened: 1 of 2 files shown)


telega-chat.el
```diff
@@ -5766,12 +5766,55 @@ Return non-nil on success."
...
+            (telega-chatbuf--highlight-quote
+             highlight
+             (button-get msg-button :value)))))
       (when callback
         (funcall callback msg-button)))
     t))
 
+(defun telega-chatbuf--remove-highlight-quote (&rest _)
+  "Remove highlighted quote."
+  (remove-overlays nil nil 'telega-chatbuf--highlight-quote t)
```
asgardthundee, line 5782: nice work

PR #465, opened 2024-03-21
add if-C-u as an option for telega-chat-send-message-on-ret

macovsky:
it would be great if you could accept this option.

I'm used to sending messages with `shift-enter` in gui client because I tend to send multi-line text mostly and to keep in mind that enter works on `(eobp)` all the time is very tedious.

thanks!

### DIFF
--- a/telega-chat.el
+++ b/telega-chat.el
@@ -4498,14 +4498,18 @@ use.  For example `C-u RET' will use
 Behaviour depends on point position and value for
 `telega-chat-send-message-on-ret'."
   (interactive)
-  (if (or (eq 'always telega-chat-send-message-on-ret)
+  (if (functionp telega-chat-send-message-on-ret)
+    (if (funcall telega-chat-send-message-on-ret)
+        (call-interactively #'telega-chatbuf-input-send)
+      (call-interactively #'newline))
+    (if (or
+          (eq 'always telega-chat-send-message-on-ret)
           (and (eobp)                   ;point at the end of the prompt
                (not (eq 'if-C-u telega-chat-send-message-on-ret)))
-          (and (or (eq 'if-at-the-end-or-C-u telega-chat-send-message-on-ret)
-                   (eq 'if-C-u telega-chat-send-message-on-ret))
+          (and (eq 'if-C-u telega-chat-send-message-on-ret)
                current-prefix-arg))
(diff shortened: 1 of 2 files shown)


zevlg:
To keep bindings you are used to you might use this configuration:
```
(define-key telega-chat-mode-map (kbd "S-<return>") #'telega-chatbuf-input-send)
(define-key telega-chat-mode-map (kbd "<return>") #'newline)
```
in your `telega-load-hook`

As to PR, it looks good!  However, I think it is better to introduce predicate as possible value for the `telega-chat-send-message-on-ret` and remove cludge `if-at-the-end-or-C-u` configuration option.  Predicate will cover any usecases

macovsky:
> To keep bindings you are used to you might use this configuration:
> 
> ```
> (define-key telega-chat-mode-map (kbd "S-<return>") #'telega-chatbuf-input-send)
> (define-key telega-chat-mode-map (kbd "<return>") #'newline)
> ```
>
> in your `telega-load-hook`

that's awesome, thank you!
 
> As to PR, it looks good! However, I think it is better to introduce predicate as possible value for the `telega-chat-send-message-on-ret` and remove cludge `if-at-the-end-or-C-u` configuration option. Predicate will cover any usecases

can you have a look? I'm not good at elisp so I might have done it badly:)

zevlg:
Code needs a little cleanup, to avoid multiple `if` and code duplication.  `if-C-u` is also should be covered by predicate setting instead, we should keep only `always`, `if-at-the-end` and predicate as value.  If you don't mind I'll add this functionality myself?

macovsky:
> Code needs a little cleanup, to avoid multiple `if` and code duplication. `if-C-u` is also should be covered by predicate setting instead, we should keep only `always`, `if-at-the-end` and predicate as value. If you don't mind I'll add this functionality myself?

makes sense! for sure I don't, it's your project.

zevlg:
Now you can `(setq telega-chat-send-message-on-ret (lambda () current-prefix-arg))` to send message only if universal argument is used with `RET`.

macovsky:
> Now you can `(setq telega-chat-send-message-on-ret (lambda () current-prefix-arg))` to send message only if universal argument is used with `RET`.

yay, thanks!

PR #468, opened 2024-04-09
Update emojis.alist to include :people-hugging:

V-Hoyle:
This is a pretty common emoji in my daily use, at least, and it'd be nice to have by default


zevlg:
Mergen in, thanks for the contribution!

PR #470, opened 2024-04-18
allow to run tests by hands from current directory

Anoncheg1:
I want to run it from the place.

### DIFF
--- a/server/run_tests.py
+++ b/server/run_tests.py
@@ -8,6 +8,8 @@ import subprocess
 TELEGA_SERVER = os.path.expanduser('~/.telega/telega-server')
 if not os.path.isfile(TELEGA_SERVER):
     TELEGA_SERVER += '.exe'
+    if not os.path.isfile(TELEGA_SERVER):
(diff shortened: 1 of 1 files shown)


server/run_tests.py
```diff
@@ -8,6 +8,8 @@
 TELEGA_SERVER = os.path.expanduser('~/.telega/telega-server')
 if not os.path.isfile(TELEGA_SERVER):
     TELEGA_SERVER += '.exe'
+    if not os.path.isfile(TELEGA_SERVER):
+        TELEGA_SERVER = "./telega-server"
```
zevlg, line 12: This won’t cover case when running tests in current directory under Windows. Better to have list of executables and check their existance before running, such as
`telega_server_bins = [‘ ~/.telega/telega-server', ‘ ~/.telega/telega-server.exe’, ‘./telega-server’, ‘./telega-server.exe’]`

zevlg (changes requested):

Issue #472, opened 2024-04-24
Telega traps on launch due to a Transient macro-expansion failure


rudolf-adamkovic:
## Telega Setup

*OS*: MacOS Sonoma 14.4.1

*Emacs*: `GNU Emacs 30.0.50 (build 1, aarch64-apple-darwin23.4.0, NS appkit-2487.50 Version 14.4 (Build 23E214)) of 2024-04-24`

*Telega*: `M-x telega-version` does not work, as it traps with the same error as `M-x telega RET`; per `M-x list-packages RET`, the version is `20240410.928`

## Steps to Reproduce

1. Launch Emacs compiled from `main`.
2. Enable MELPA and install Telega using =package-install=.
3. `M-x telega RET`.

## Curent Behavior

```
internal-macroexpand-for-load: Eager macro-expansion failure: (error "transient-define-*: Interactive form missing")
```

## Expected Behavior

Running Telega.

zevlg:
What is your `transient` version ?

cnsunyour:
Same issue. 
It seems to be caused by https://github.com/magit/transient/commit/7e827c31d63e6aaff6511c33f436dbcba619de4e

rudolf-adamkovic:
> What is your `transient` version ?

```
Package transient is built-in.

     Status: Built-In.
    Version: 0.6.0
    Summary: Transient commands
Required by: telega-20240410.928, magit-20240423.2030, git-commit-20240415.1544,
               geiser-guile-20240314.1950, ement-0.15, cider-20240423.1541
Other versions: 20240421.1737 (installed), 20240421.1737 (melpa), 0.6.0 (melpa-stable), 0.6.0 (gnu).
```

PR #473, opened 2024-04-26
[fix] add missing interactive to transient-define-prefix

HyperSuperMetaCtrl:
without this telega does not start anymore
Addresses Issue #472

### DIFF
--- a/telega-util.el
+++ b/telega-util.el
@@ -3069,6 +3069,7 @@ Return nil if there is no tags for the SM-TOPIC-ID or new tag is choosen."
 ;; scope for this commands is cons cell where car is tag and cdr is
 ;; msg
 (transient-define-prefix telega-saved-messages-tag-commands (tag)
+  (interactive)
   [:description
    (lambda ()
      (telega-ins--as-string
--- a/telega-util.el
+++ b/telega-util.el
@@ -3069,7 +3069,6 @@ Return nil if there is no tags for the SM-TOPIC-ID or new tag is choosen."
 ;; scope for this commands is cons cell where car is tag and cdr is
 ;; msg
 (transient-define-prefix telega-saved-messages-tag-commands (tag)
-  (interactive)
   [:description
    (lambda ()
      (telega-ins--as-string
@@ -3079,11 +3078,7 @@ Return nil if there is no tags for the SM-TOPIC-ID or new tag is choosen."
    (telega-saved-messages-tag-filter)
    (telega-saved-messages-tag-add-name)
    (telega-saved-messages-tag-remove)
-   ]
-  (lambda (tag)
-    (interactive (list (user-error "Do not call this command directly")))
-    ;; (transient-setup 'telega-saved-messages-tag-commands nil nil :scope (cons tag msg))
-    ))
+   ])
 
 
 ;;; Stipple drawing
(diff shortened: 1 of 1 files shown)


HyperSuperMetaCtrl:
Starting telega with `M-x` returned:
```
internal-macroexpand-for-load: Eager macro-expansion failure: (error "transient-define-*: Interactive form missing")
```
I then turned on `tooggle-debug-on-error` and got:
```
Debugger entered--Lisp error: (error "Eager macro-expansion failure: (error \"transient-d...")
  error("Eager macro-expansion failure: %S" (error "transient-define-*: Interactive form missing"))
  internal-macroexpand-for-load((transient-define-prefix telega-saved-messages-tag-commands (tag) [:description (lambda nil (telega-ins--as-string (telega-ins--saved-messages-tag (car (oref transient--prefix scope))) (telega-ins--with-face 'transient-heading (telega-ins "Tag Commands")))) (telega-saved-messages-tag-filter) (telega-saved-messages-tag-add-name) (telega-saved-messages-tag-remove)] (lambda (tag) (interactive (list (user-error "Do not call this command directly"))))) nil)
  load-with-code-conversion("/home/me/.config/emacs/.local/straight/build-2..." "/home/me/.config/emacs/.local/straight/build-2..." nil t)
  require(telega-util)
  load-with-code-conversion("/home/me/.config/emacs/.local/straight/build-2..." "/home/me/.config/emacs/.local/straight/build-2..." nil t)
  require(telega-root)
  load-with-code-conversion("/home/me/.config/emacs/.local/straight/build-2..." "/home/me/.config/emacs/.local/straight/build-2..." nil t)
  command-execute(telega record)
  execute-extended-command(nil "telega" nil)
  funcall-interactively(execute-extended-command nil "telega" nil)
  command-execute(execute-extended-command)
```
Then added `(interactive)` to `transient-define-prefix` in `telega-util.el`.

With this change telega starts as usual.

rudolf-adamkovic:
I can confirm that this works.

Thanks @HyperSuperMetaCtrl for figuring out the problem!

HyperSuperMetaCtrl:
Happy to help

telega-util.el
```diff
@@ -3069,6 +3069,7 @@ Return nil if there is no tags for the SM-TOPIC-ID or new tag is choosen."
 ;; scope for this commands is cons cell where car is tag and cdr is
 ;; msg
 (transient-define-prefix telega-saved-messages-tag-commands (tag)
+  (interactive)
```
zevlg, line 3072: I think there should be a docstring for this transient prefix, not `(interactive)`.  `telega-saved-messages-tag-commands` is not supposed to be called interactive

zevlg (changes requested):

telega-util.el
HyperSuperMetaCtrl: Hey zevlg,
thank you for your feedback.
What do you mean exactly? Should I add a docstring, that warns about not using this function interactively, or do you want it to be solved in a different way, because this transient prefix is not supposed to be called interactively?

zevlg: In the place where you've added `(interactive)` should be a docstring for the transient prefix.  `(interactive)` works because it adds missing argument and docstrings are ignored.  So just replace `(interactive)` with the docstring

HyperSuperMetaCtrl: Just replacing the `(interactive)` with a string leads to the same error as before for me. But my elisp is quite limited. So this could just be a skill issue on my side

HyperSuperMetaCtrl:
Found a way to do this without adding `(interactive)`
closing

PR #474, opened 2024-04-28
[fix] error "transient-define-*: Interactive form missing"

### LINKED ISSUE #472: Telega traps on launch due to a Transient macro-expansion failure

HyperSuperMetaCtrl:
without this telega does not start anymore
fixes #472

In contrast to #473  this does not make concerning function `(interactive)` and instead just add `[]`.

### DIFF
--- a/telega-util.el
+++ b/telega-util.el
@@ -3079,10 +3079,10 @@ Return nil if there is no tags for the SM-TOPIC-ID or new tag is choosen."
    (telega-saved-messages-tag-add-name)
    (telega-saved-messages-tag-remove)
    ]
-  (lambda (tag)
+  [(lambda (tag)
     (interactive (list (user-error "Do not call this command directly")))
     ;; (transient-setup 'telega-saved-messages-tag-commands nil nil :scope (cons tag msg))
-    ))
+    )])
 
 
 ;;; Stipple drawing


HyperSuperMetaCtrl:
I think this still does not really fix the underlying problem

zevlg:
BTW: what is your `transient` version triggering the problem?

HyperSuperMetaCtrl:
I think its `0.6.0`

benthamite:
I'm personally experiencing this with the latest `transient` and `telega` commits, as of now: `2dd0102` and `d73e02d` respectively.

jessy836:
I have same problem. I pined transient to commit   "8a80e95" and now it's work again. Not tested with most recent commit.

jessy836:
I tryed to pin transient to different commit and see what hapend, here are the resultat.

- fcc60e2 work ; if I understand well is the most recent commit that work at this moment 
- 7e827c3 does not work; If i understand well is the first commit that triger the error 
- 2dd0102  (last commit available at this moment) does not work 

I don't tested commit between 7e827c3 and  2dd0102.
Hop that this can help !

licht1stein:
I have the same problem. Tried setting up telega on a new emacs setup.

PR #484, opened 2024-06-14
Use repology badge for Guix package.

sarg:
Use repology latest version badge to avoid manual bumping


zevlg:
Merged in, thanks!

PR #487, opened 2024-08-10
Compatible with podman 5.2.0, check device existence.

BlindingDark:
In Podman 5.2.0, adding non-existent devices will throw an error.

> The --device option to podman create and podman run is no longer ignored when --privileged is also specified (https://github.com/containers/podman/issues/23132).

This PR will check if the device exists when add host devices to container.


zevlg:
Merged in, thanks!

PR #495, opened 2024-09-11
[fix] Dockerfile: set tdlib version by ARG

BlindingDark:
This will get us a stable version of tdlib when we rebuild.


zevlg:
Merged in, thanks!

PR #497, opened 2024-10-04
fix forwarding

haruhi-s:
Currently the flag `can_be_forwarded` is checked upon forwarding messages in a chat, but it is discarded in the latest version of TDLib, causing forwarding messages to be impossible. Checking `can_be_saved` is equivalent.


zevlg:
This patch is incorrect.  Use TDLib 1.8.29 for now.  Be patient, support for newest TDLib will be added eventually

PR #501, opened 2024-11-01
set current version, add git sha

macovsky:


docs/telega-ellit.org
```diff
@@ -111,8 +111,8 @@ Thanks to [[https://t.me/ritsch_master][=@ritsch_master=]]
 ** Building TDLib
 
 [[https://core.telegram.org/tdlib][TDLib]] is the library for building Telegram clients. It requires a
-large amount of memory to be built.  Make sure you are using TDLib
-version greater or equal to {{{eval(telega-tdlib-min-version, t)}}}.
+large amount of memory to be built.  Make sure you are using TDLib version
```
macovsky, line 114: WDYT about setting the concrete version instead of `greater or equal to` since 1.8.38 is not supported yet?

also added SHA to help with installation

macovsky: also I can add installation instructions for archlinux yay if you are interested (if that SHA thing will go through)

telega.el
```diff
@@ -12,6 +12,7 @@
 (defconst telega-version "0.8.293")
 (defconst telega-server-min-version "0.7.7")
 (defconst telega-tdlib-min-version "1.8.29")
+(defconst telega-tdlib-min-version-sha "12c1689b5" "SHA for minimum version. Used for manual generation.")
```
zevlg, line 15: Maybe add sha1 as part of `telega-tdlib-min-version` and strip it in the `telega--on-updateOption` while validating version?

i.e. `(defconst telega-tdlib-min-version "1.8.29-12c1689b5")`, it will be easier to maintain

docs/telega-ellit.org
macovsky: what do you think of using a list instead? that will allow to keep the code intact.

telega.el
macovsky: what do you think about using a list instead? that will allow to keep the code intact.

docs/telega-ellit.org
zevlg: Too many variables, simple string like "1.8.39-056963e" is better, also, it resembles what `M-x telega-version RET` returns

macovsky: how about a compromise: one extra var as simple string for manual but code stays intact?

telega.el
```diff
@@ -11,7 +11,8 @@
 ;; Version: 0.8.293
 (defconst telega-version "0.8.293")
 (defconst telega-server-min-version "0.7.7")
-(defconst telega-tdlib-min-version "1.8.29")
+(defconst telega-tdlib-min-version-with-sha "1.8.29-12c1689b5")
+(defconst telega-tdlib-min-version (car (split-string telega-tdlib-min-version-with-sha "-")))
```
zevlg, line 15: Wait, we have sha1 in the `etc/Dockerfile`, let's extract it from there while generating docs!

This avoids adding sha1 into multiple different places

PR #502, opened 2024-11-06
add `telega-chat-skip-delete-confirmation-for` custom option

macovsky:
specifies Chat Temex for chats which can be deleted without typing confirmation

### DIFF
--- a/docs/telega-ellit.org
+++ b/docs/telega-ellit.org
@@ -573,6 +573,7 @@ Use ~telega-story-match-p~ to match a story.
 - {{{user-option(telega-chat-group-messages-for, 2)}}}
 - {{{user-option(telega-chat-show-deleted-messages-for, 2)}}}
 - {{{user-option(telega-root-view-grouping-alist, 2)}}}
+- {{{user-option(telega-chat-delete-skip-confirmation-for, 2)}}}
 
 * Chat Folders
 
--- a/telega-customize.el
+++ b/telega-customize.el
@@ -215,6 +215,13 @@ Tracking notifications for telega buffers will use the
              (or unmuted mention))
   :group 'telega)
 
+(defcustom telega-chat-delete-skip-confirmation-for nil
+  "*Specifies Chat Temex for chats which can be deleted without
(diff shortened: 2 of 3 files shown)


telega-customize.el
```diff
@@ -215,6 +215,13 @@ Tracking notifications for telega buffers will use the
              (or unmuted mention))
   :group 'telega)
 
+(defcustom telega-skip-delete-confirmation-for nil
+  "*Specifies Chat Temex for chats which can be deleted without
```
macovsky, line 219: btw what this `*` is about? I just copied it from `telega-use-tracking-for`

docs/telega-ellit.org
```diff
@@ -573,6 +573,7 @@ Use ~telega-story-match-p~ to match a story.
 - {{{user-option(telega-chat-group-messages-for, 2)}}}
 - {{{user-option(telega-chat-show-deleted-messages-for, 2)}}}
 - {{{user-option(telega-root-view-grouping-alist, 2)}}}
+- {{{user-option(telega-skip-delete-confirmation-for, 2)}}}
```
macovsky, line 576: do I have to generate the manual?

telega-customize.el
zevlg: "*" Stands for "user might want to change default value".

Name it like `telega-chat-delete-skip-confirmation-for` (i.e. "telega-chat-" prefix matters) and put this custom option to `telega-chat` customization group.

zevlg (changes requested):

docs/telega-ellit.org
zevlg: > do I have to generate the manual?

No need, I'll rebuild docs

telega-customize.el
macovsky: awesome, fixed!

zevlg:
Merged in, thank you for contribution!

macovsky:
> Merged in, thank you for contribution!

excellence! thanks for your thoughtful consideration as always—this new custom option is much better!

PR #512, opened 2025-01-24
update Dockerfile alpine:latest

eins:
update alpine from 3.13 to latest (3.21)
replace libappindicator-dev for ayatana-appindicator


zevlg:

Issue #519, opened 2025-03-22
Empty file path when saving a file with downloading


Vonfry:
**Reproduce**
1. find a message with a file that need to be download (no cached).
2. call `telega-msg-save`
3. After downloaded, it will prompt a path, which is the same as `telega-msg-save-dir`,already existing

**Trace**
The issue is caused at line 47 in `telega-msg-save`, where `(telega--tl-get file :local :path)` returns an empty string, and then the condition `(and fname telega-msg-save-dir)` is passed, so `new-fpath` is equal to `telega-msg-save-dir` directly though `(expand-file-name fname telega-msg-save-dir)`.

I think fpath here should be nil, check fname with empty string testing, or set that attrubitues correctly after downloaded.

**Emacs and telega version**
Emacs: `GNU Emacs 31.0.50 (build 1, x86_64-pc-linux-gnu, X toolkit, cairo version 1.18.2, Xaw3d scroll bars)`

Telega: `telega v0.8.452 (TDLib v1.8.46-GITDIR-) (telega-server v0.8.3)`, which should be at date `20250304.1550`

PR #524, opened 2025-07-20
fix wrong path handle on msg save when downloading file

Vonfry:
fixes: #519 

`telega-file--download` is async and `telega-msg-save` doesn't handle the async process correctly, IMO.


zevlg:
Thanks for the patch, however, it does what is not intended.  What we want is to start downloading file, and while file is downloading we query user to read a filename to save to.

See comment in the code
```
;; NOTE: Start downloading file in the background while reading
;; filename
```

PR #528, opened 2025-09-15
fix  file reference and time of getting properties 

Vonfry:
Thank for your patches on #519. But it doesn't really fix the issue.

fname and new-fpath are get from file, before finished download, where they are always fetched wrong values, especially on fname that is always empty.
They should be set in callback functions to get the correct value which are set in `telega-file--download`.
I recreate pr #524 without conflictions.

### DIFF
--- a/telega-msg.el
+++ b/telega-msg.el
@@ -1704,39 +1704,31 @@ Saved Messages."
 
       ;; NOTE: Start downloading file in the background while reading
       ;; filename
-      (unless (telega-file--downloaded-p file)
-        (telega-file--download file
-          :priority 32
-          :update-callback
-          (lambda (_dfile)
-            (telega-msg-redisplay msg))))
-
-      (let* ((fname (file-name-nondirectory
-                     (telega--tl-get file :local :path)))
-             (new-fpath
-              (if (and telega-msg-save-dir (not (string-empty-p fname)))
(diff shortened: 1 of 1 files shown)


zevlg:
> fname and new-fpath are get from file, before finished download, where they are always fetched wrong values, especially on fname that is always empty.

What do you mean by "wrong values"?

If filename is not available, then filename is read from user using `default-directory` directory.  You can customize `telega-msg-save-dir` to be your default-directory to download files to

zevlg:
> Furthermore, line 1721 use fname as INITIAL for read-file-name, which is empty as well if file hasn't been downloaded already. It is the same reason why (and telega-msg-save-dir (not (string-empty-p fname))) cannot be passed. I know the propose is to ask filename when downloading. However if not in update-callback, neither saving in telega-msg-save-dir without asking or providing a default filename in prompt can work. 

I don't understand, why it can't work? It works right now.  Just specify filename you want to save file to, that's it.

Probably you don't want to specify filename explicitly, in this case first download the file by pressing [Download] button, and then save it, in this case you will get filename provided by Telegram.

File can be very large, waiting for it being downloaded might take a time, we can't query user input in the callback when downloading is completed, this is bad, unpredicted UI

Vonfry:
Thanks!

PR #536, opened 2026-01-21
Support pasting multiple files to chatbuf

sarg:
This allows `Ctrl-C`-ing multiple files in a GUI file manager and then `C-c C-v` them into a chatbuf. Tested in `thunar` and `pcmanfm` under X11.


aragaer:
`(gui-get-selection 'CLIPBOARD 'text/uri-list)` returns binary garbage when clipboard contents is an image. This commit breaks the feature where `C-c C-v` would just paste an image from clipboard directly.

PR #539, opened 2026-01-30
don't add non-existent contrib to load-path

sarg:
in guix contrib is packaged separately, which leads to non-existent dir added to load-path by telega


zevlg:
Is this PR still needed?  sorry for the delay

sarg:
I've fixed it in guix already: https://codeberg.org/guix/guix/pulls/5571
up to you if you want it merged

telega.el
```diff
@@ -386,7 +386,9 @@ string at point."
 (provide 'telega)
```
zevlg, line 388: Could you please add comment here to describe why we check "contrib" for existance

zevlg (changes requested):

telega.el
sarg: done

zevlg:
Merged in, thank you!

PR #542, opened 2026-02-13
Fix image pasting from clipboard

aragaer:
#536 broke the ability to paste images directly to chatbuf. This change adds a check for clipboard contents type to see if it's text or binary data.

### DIFF
--- a/telega-chat.el
+++ b/telega-chat.el
@@ -5858,7 +5858,10 @@ If `\\[universal-argument]' is given, then attach clipboard as document."
           (error "No image in CLIPBOARD"))
         (telega-chatbuf-attach-media tmpfile (when as-file-p 'preview)))
 
-    (if-let* ((urls (gui-get-selection 'CLIPBOARD 'text/uri-list)))
+    (if-let* ((urls (and (cl-position "text/plain"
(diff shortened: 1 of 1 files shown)


telega-chat.el
```diff
@@ -5858,7 +5858,10 @@ If `\\[universal-argument]' is given, then attach clipboard as document."
           (error "No image in CLIPBOARD"))
         (telega-chatbuf-attach-media tmpfile (when as-file-p 'preview)))
 
-    (if-let* ((urls (gui-get-selection 'CLIPBOARD 'text/uri-list)))
+    (if-let* ((urls (and (cl-position "text/plain"
```
sarg: I assumed `gui-get-selection` would return `nil` when the TARGET is unsupported by the selection owner (as it kind of should respond). However, it seems `xclip` deviates from the conventions responding with image data when unsupported `text/uri-list` target is requested. 

Now to make it foolproof, you can check TARGETs first and request `text/uri-list` only when it is advertised. That means you need to check for `text/uri-list` presence and not for `text/plain`.

Please test afterwards with copy-pasting from a file manager, e.g. pcmanfm or thunar.

See: https://tronche.com/gui/x/icccm/sec-2.html

PR #550, opened 2026-02-26
telega-bridge-bot: fix Matrix user avatar fetching

LuciusChen:
Several interrelated issues prevented Matrix bridge bot avatars from loading correctly in certain rooms:

- URL-encode Matrix room IDs in joined_members request: room IDs contain '!' and ':' which caused Emacs url-retrieve to silently fail without invoking the callback.

- Deduplicate concurrent room member fetches: multiple messages rendered simultaneously for the same room triggered parallel url-retrieve calls that exceeded Emacs's per-host connection limit, causing fetches to hang.  Use a callbacks hash table to queue duplicate requests and fire them all when the single fetch completes.

- Cache members by MXID in addition to display name: some bridges format messages as "@user:server.org: body", so looking up the avatar by display name missed these entries.

- Always invoke download-async callback: the callback was only called inside the 200-OK branch, so a 404 thumbnail response silently swallowed the error and never triggered the fallback download.

- Use authenticated media endpoint as fallback: homeservers implementing MSC3916 (Matrix 1.11 authenticated media, e.g. mozilla.org, magdeburg.jetzt) reject unauthenticated requests. Replace the direct-server fallback with /_matrix/client/v1/media/download/ using an Authorization: Bearer header so the homeserver federates the request on our behalf.

- Strip t2bot status-emoji prefix from usernames: t2bot prepends a Unicode status indicator (e.g. "🟢 ") to display names, causing the avatar cache lookup to miss the stored entry.

### DIFF
--- a/contrib/telega-bridge-bot.el
+++ b/contrib/telega-bridge-bot.el
@@ -106,6 +106,8 @@
 
 (defvar telega-bridge-bot--matrix-room-cache nil)
 (defvar telega-bridge-bot--matrix-room-cache-last-modified-plist '())
+(defvar telega-bridge-bot--matrix-room-fetch-callbacks (make-hash-table :test 'equal)
+  "Maps room-id to list of pending callbacks waiting for members fetch.")
 
 (defvar telega-bridge-bot--matrix-host "https://matrix-client.matrix.org")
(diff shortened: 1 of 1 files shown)


zevlg:
Merged in, thank you!

PR #552, opened 2026-02-27
contrib: fix matrix bridge name splitting for :: display names

LuciusChen:
## Summary

  This patch fixes a parsing bug in `contrib/telega-bridge-bot.el` for Matrix-bridged text messages.

  Previously, `telega-bridge-bot--matrix-text-spliter` used a naive split on `: `, which incorrectly broke display names containing `::` (for example, `terrorist :: Feline: ...`), causing part of the name to appear in the
  message body.

  ## What changed

  - Updated `telega-bridge-bot--matrix-text-spliter` to:
      - find the first ": " delimiter that is not part of `::` ;
      - preserve the full display name when `::` appears in the name;
      - return `(text, "")` when no valid delimiter is found (safe fallback).
  - Kept existing `matterbridge` special-case behavior unchanged.

  ## Example

  Before:

  - Input: `terrorist :: Feline: moto ...`
  - Parsed name: `terrorist :`
  - Parsed body: `Feline: moto ...`

  After:

  - Parsed name: `terrorist :: Feline`
  - Parsed body: `moto ...`

### DIFF
--- a/contrib/telega-bridge-bot.el
+++ b/contrib/telega-bridge-bot.el
@@ -287,16 +287,28 @@ If FORCE-UPDATE is non-nil, force update the file."
 
 (defun telega-bridge-bot--matrix-text-spliter (text)
   "Split TEXT into username and message."
-  (let* ((name-and-body (split-string text ": "))
-         (name (car name-and-body))
-         (body (string-join (cdr name-and-body) ": ")))
+  (let ((search-pos 0)
+        (split-pos nil))
+    ;; Find first `: ' that is not part of `:: ' in display name.
+    (while (and (not split-pos)
+                (setq search-pos (string-match ": " text search-pos)))
+      (unless (and (> search-pos 0)
+                   (eq (aref text (1- search-pos)) ?:))
+        (setq split-pos search-pos))
+      (setq search-pos (1+ search-pos)))
+    (let* ((name (if split-pos
+                     (substring text 0 split-pos)
+                   text))
+           (body (if split-pos
+                     (substring text (+ split-pos 2))
+                   "")))
     (if (and (string-equal name "matterbridge")
              (string-match "^\\[\\(.*?\\)\\]" body))
         ;; get the username in []
         (list
          (concat (match-string 1 body) " <matterbridge>") ; username
          (substring body (+ (match-end 0) 1))) ; message
-      (list name body))))
+      (list name body)))))
 
 (defun telega-bridge-bot--matrix-file-spliter (text)
   "Split TEXT into username and message."


contrib/telega-bridge-bot.el
```diff
@@ -287,16 +287,28 @@ If FORCE-UPDATE is non-nil, force update the file."
...
+      (unless (and (> search-pos 0)
+                   (eq (aref text (1- search-pos)) ?:))
+        (setq split-pos search-pos))
+      (setq search-pos (1+ search-pos)))
+    (let* ((name (if split-pos
+                     (substring text 0 split-pos)
+                   text))
+           (body (if split-pos
+                     (substring text (+ split-pos 2))
+                   "")))
```
zevlg: Are you sure you want this pretty complex code?  Maybe just splitting by "[^:]: " instead of ": " will work?

LuciusChen:
How about this？

```diff
diff --git a/contrib/telega-bridge-bot.el b/contrib/telega-bridge-bot.el
--- a/contrib/telega-bridge-bot.el
+++ b/contrib/telega-bridge-bot.el
@@ -287,10 +287,15 @@ If FORCE-UPDATE is non-nil, force update the file."
 (defun telega-bridge-bot--matrix-text-spliter (text)
   "Split TEXT into username and message."
-  (let* ((name-and-body (split-string text ": "))
-         (name (car name-and-body))
-         (body (string-join (cdr name-and-body) ": ")))
+  (let* ((split-pos
+          ;; Find first ": " that is not part of ":: " in display name.
+          ;; Group 1 captures only the actual delimiter, so we can split
+          ;; without losing the preceding character.
+          (when (string-match "\\(?:\\`\\|[^:]\\)\\(: \\)" text)
+            (match-beginning 1)))
+         (name (if split-pos (substring text 0 split-pos) text))
+         (body (if split-pos (substring text (+ split-pos 2)) "")))
     (if (and (string-equal name "matterbridge")
              (string-match "^\\[\\(.*?\\)\\]" body))
         ;; get the username in []
         (list
          (concat (match-string 1 body) " <matterbridge>") ; username
          (substring body (+ (match-end 0) 1))) ; message
       (list name body))))
```

`[^:]`:  cannot be used with `split-string` here because `split-string` removes the full matched separator.
  In that regex, `[^:]` is part of the match, so it also gets removed, which drops the last character of the username.

  Example:

  - Input: `Alice: hi`
  - `split-string` with `[^:]`:  gives `("Alic" "hi")` (loses `e`)

  So `[^:]` is fine for detecting context, but not as a separator for `split-string` unless you keep the separator position via `string-match` and split manually.

zevlg:
Ah wait, you did not changed actual code? And I've merged in %)

LuciusChen:
Should I synchronize the code, make revisions, and submit another PR, or would you prefer to manually merge the patch above into the master branch?

PR #555, opened 2026-03-04
Fix addProxy param

phanen:
- **Bump msv to 1.8.61**
- **Fix addProxy param**

`telega-proxies` don't work for me about 1 month.
After digging it seems to be upstream change https://github.com/tdlib/td/commit/7a0e70b161bade9fa35a515abd5a76b7210ff0d3.

I have to bump the min support version since it can break telega-proxies on <1.8.61.


zevlg:
Merged in, thank you!

PR #566, opened 2026-04-07
fix getting version of telega-server

Anoncheg1:
Instead of using `shell-command` we use `process-lines` that treat command and argument separately, this dont break security systems, if they are used for Emacs instance. I require this. No difference for you.

Without this I get wrong version error.

### DIFF
--- a/telega-server.el
+++ b/telega-server.el
@@ -188,8 +188,8 @@ Raise error if not found."
 
 (defun telega-server-version ()
   "Return telega-server version."
-  (let ((ts-usage (shell-command-to-string
-                   (telega-server--process-command "-h"))))
+  (let ((ts-usage (car (process-lines
+                        (telega-server--process-command) "-h"))))
     (when (string-match "^Version \\([0-9.]+\\)" ts-usage)
       (match-string 1 ts-usage))))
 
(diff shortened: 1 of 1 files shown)


zevlg:
Please elaborate on security concerns you have mentioned.  What does `M-: (telega-server-version) RET` returns to you and what Emacs version do you use?

Changes you propose will break docker support, i.e. `(telega-server-version)` will fail if `telega-use-docker` is non-nil

Anoncheg1:
(telega-server-version) return nothing, because I run Emacs without shell support.
I notices this Docker shit inside of this file.
I guess I found reason why software getting old so fast and AI fast vibe condig is so popular - because some programmers dont know that app should be environment-agnostic, respecting the interface. it should not support everything and everyone It should be simple and minimal (KISS). Here without shell support I need.

zevlg:
Telega uses various shell commands all over the place to do things.  If running shell commands is insecure on your system it is better not to run telega there.

Free software is free, you are free to use it, not to use it, modify it to fit your needs, expectations and vision about what is "right thing"

Anoncheg1:
> If running shell commands is insecure on your system it is better not to run telega there.

Only one function break Telega - `telega-server-version`, fixed here. It is not about to use shell or not.

 It is about which way you call programs- 

**secure with separate parameters**  without shell

OR

 **insecure by sending one line into shell**. 

And whether you require Docker and other strange "features".

zevlg:
It is okay not to use shell unless it breaks existing functionality.  This PR breaks it, thats why I can't accept it.  Please fix it to keep existing functionality

PR #567, opened 2026-04-10
[codex] Fix forum topic lookup from topic messages

liaowang11:
## What changed

- Added `telega-msg-topic-ensure` to resolve `messageTopicForum` messages into cached `forumTopic` objects when the topic info has not been loaded yet.
- Switched topic opening and topic info display to use that resolver instead of assuming the topic is already cached.
- Added focused ERT coverage for bot chats with `has_topics` and for opening a forum-topic message when the topic cache is cold.

## Why

Bot private-chat topics in TDLib are represented through the existing forum-topic path. telega already marks bots with `has_topics` as forum-capable, but opening a topic from a message still assumed the `forumTopic` object had already been cached. That broke topic navigation on cache misses and would also block the bot-private-topic flow.

## Impact

- Topic messages can now resolve and open their topic lazily instead of failing when topic metadata was not preloaded.
- The fix improves existing forum-topic behavior and covers bot private chats that expose topics through the same TDLib model.

## Validation

- `emacs -Q --batch -L . -l test.el -f ert-run-tests-batch-and-exit`


telega-topic.el
```diff
@@ -234,6 +236,16 @@
     (telega-topic-get (telega-msg-chat msg)
                       (telega--MessageTopic-id msg-topic))))
 
+(defun telega-msg-topic-ensure (msg)
```
zevlg: Probably better to add `sync-p` param to `telega-msg-topic`, to return message's topic in sync manner, i.e. fetch it in case it is not yet in the cache.

telega-topic.el
```diff
@@ -234,6 +236,16 @@
...
+  "Return topic for the message MSG, fetching it if needed."
+  (or (telega-msg-topic msg)
+      (when-let* ((msg-topic (plist-get msg :topic_id))
+                  ((eq (telega--tl-type msg-topic) 'messageTopicForum))
```
zevlg: use `(when-let ((topic (telega-msg-match-p msg 'is-forum-topic))` istread, it returns topic only if message's topic is forum topic

telega-topic.el
```diff
@@ -304,7 +316,9 @@ If START-MSG-ID is specified, jump to the this message in the topic."
 (defun telega-msg-show-topic-info (msg)
   "Show MSG's topic info."
   (interactive (list (telega-msg-for-interactive)))
-  (telega-describe-topic (telega-msg-topic msg)))
+  (if-let ((topic (telega-msg-topic-ensure msg)))
```
zevlg: If flag for `telega-msg-topic` is added, then only change will be like `(telega-msg-topic msg 'sync)`

zevlg (changes requested):

test-topics.el
```diff
@@ -0,0 +1,71 @@
+;;; test-topics.el --- Topic tests for telega  -*- lexical-binding:t -*-
```
zevlg: Why having separate file to test topics? Why not to put all these tests into test.el ?

liaowang11:
Sorry for the mistaken force-push, I changed according to you review. 
I'm not quite sure if it's to your liking to add error guards, if not, I will delete those.

zevlg:
> Sorry for the mistaken force-push, I changed according to you review. I'm not quite sure if it's to your liking to add error guards, if not, I will delete those.

No, errors guards are non needed, this is an assertion, you can add `cl-assert` calls to catch internal errors in the telega or TDLib

telega-topic.el
```diff
@@ -228,11 +230,18 @@
...
-    (telega-topic-get (telega-msg-chat msg)
-                      (telega--MessageTopic-id msg-topic))))
+(defun telega-msg-topic (msg &optional sync-p)
+  "Return topic for the message MSG.
+If SYNC-P is non-nil, fetch missing forum topic info before returning it."
+  (when-let ((msg-topic (plist-get msg :topic_id))
+             (chat (telega-msg-chat msg 'offline)))
+    (or (telega-topic-get chat (telega--MessageTopic-id msg-topic))
+        (when-let* (sync-p
+                    (forum-topic-msg (telega-msg-match-p msg 'is-forum-topic))
```
zevlg: Please use `msg-topic` or `msg-forum-topic` name, "msg" at the end of the name is a bit missleading

telega-topic.el
```diff
@@ -228,11 +230,18 @@
...
+                    (topic (telega--getForumTopic
+                               chat (plist-get forum-topic-msg :forum_topic_id))))
+          (unless (telega--tl-error-p topic)
```
zevlg: Probably here is the best place to assert that there is no error after getting topic. If some topic is declared in the msg and error is returned, then there is definitly some internal error in the telega or TDLib. Having assert here allows to remove error checks in the code that uses `sync-p`

zevlg (changes requested):

telega-topic.el
liaowang11: sure, changed to `msg-forum-topic`

zevlg:
Code is ready to be merged, but test fails

Take a look at test failure - https://github.com/zevlg/telega.el/actions/runs/24309991722/job/71060459617?pr=567

liaowang11:
> Code is ready to be merged, but test fails
> 
> Take a look at test failure - https://github.com/zevlg/telega.el/actions/runs/24309991722/job/71060459617?pr=567

Would you please run the test again?

zevlg:
Merged in! Thank you for your contribution

PR #569, opened 2026-04-14
fix(completions): init emoji once in CAPF setup

liaowang11:
## Summary

The `re-search-backward` in the emoji CAPF functions uses `telega-emoji-candidate-max-length` as the search bound:

```elisp
(re-search-backward "\\(?:^\\|[[:space:]]\\)\\(:[^: _]+\\)"
  (max (point-min) (- end telega-emoji-candidate-max-length)) t)
```

When `telega-emoji-candidate-max-length` is 0 (before init), the bound collapses to `end`, leaving no range to search backward — so the regex never matches, the `when-let*` falls through, and `telega-emoji-init` (previously inside the match block) is never reached.

Move the init call to `telega-completions-setup-capf`, where it runs once at chat buffer setup time, only when an emoji CAPF function is present in `telega-completions-capf-functions`.

## Test plan

- [x] All 20 existing tests pass including new emoji CAPF tests
- [x] CI passes on both Emacs 28.1 and snapshot
- [x] Verify emoji completion works in a chat buffer with CAPF
- [x] Verify emoji completion still works with company backend

### DIFF
--- a/telega-completions.el
+++ b/telega-completions.el
@@ -238,7 +238,6 @@ If FUZZY-MATCH is non-nil, also allow fuzzy hyphen matching."
 (defun telega-completions--emoji-candidates (prefix fuzzy-match)
   "Return local emoji candidates matching PREFIX.
 Use FUZZY-MATCH to control fuzzy hyphen matching."
-  (telega-emoji-init)
   (cl-remove-if-not
    (lambda (emoji-name)
      (telega-completions--emoji-match-p prefix emoji-name fuzzy-match))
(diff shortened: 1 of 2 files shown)


telega-completions.el
```diff
@@ -583,7 +582,6 @@ Handles repeated leading CHARs (e.g. @@).  Returns nil if not applicable."
                             (match-beginning 1))))
               (prefix (buffer-substring-no-properties start end))
               ((string-prefix-p ":" prefix)))
-    (telega-emoji-init)
```
zevlg, line 586: Just move `(telega-emoji-init)` above `when` clause so emojis will be initialized before variable is used.

telega-completions.el
```diff
@@ -771,6 +769,9 @@ Handles repeated leading CHARs (e.g. @@).  Returns nil if not applicable."
 (defun telega-completions-setup-capf ()
   "Add telega CAPF functions to `completion-at-point-functions'.
 Intended for use in `telega-chat-mode-hook'."
+  (when (or (memq 'telega-capf-emoji telega-completions-capf-functions)
```
zevlg: User might not use `telega-completions-setup-capf` and configure `completion-at-point-functions` directly in the `telega-chat-mode-hook`, so `telega-emoji-init` will not be called at all

zevlg (changes requested):

zevlg:
Merged in, thank you!

PR #571, opened 2026-04-29
telega-server: run gtk_init_check in the appindicator loop thread

akaWolf:
GTK's documented threading rules require GTK initialization, the GMainLoop driving the events, and any GTK / libappindicator calls to all happen on the same thread (see "GTK+ is not thread safe, and you should only use GTK+ and GDK from the thread gtk_init() and gtk_main() were called on", https://docs.gtk.org/glib/struct.MainContext.html and the GTK threading docs).

Previously gtk_init_check() ran in the main thread while telega_appindicator_loop() ran g_main_loop_run() in a separate pthread, splitting initialization and event dispatch across two threads.  As a result, idle callbacks scheduled from the main thread by telega_appindicator_send() via g_idle_add() (which targets the global default GMainContext that GTK has set up) were not dispatched by the loop, so "setup", "status", "icon" and "label" commands were silently dropped and no tray icon was ever shown.

Move gtk_init_check() into telega_appindicator_loop() so initialization and the main loop both live on the same thread, matching GTK's threading model.

### DIFF
--- a/server/telega-appindicator.c
+++ b/server/telega-appindicator.c
@@ -31,7 +31,6 @@
  */
 #include <stdio.h>
 #include <stdlib.h>
-#include <stdbool.h>
 #include <assert.h>
 
 #ifdef WITH_AYATANA_APPINDICATOR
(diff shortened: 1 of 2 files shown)


server/telega-appindicator.c
```diff
@@ -137,12 +137,23 @@ telega_appindicator_send(const char* json)
...
+        return true;
 }
 
 void*
 telega_appindicator_loop(void* data)
 {
+        if (!gtk_init_check(NULL, NULL)) {
+                fprintf(stderr, "[telega-appindicator] gtk_init_check failed\n");
+                return NULL;
+        }
```
zevlg, line 148: This is not a failure.  User might run telega-server with appindicator code in the tty, so gtk can't be initialized.  Better to write something like "appindicator disabled"

server/telega-appindicator.c
```diff
@@ -137,12 +137,23 @@ telega_appindicator_send(const char* json)
...
 {
-        return gtk_init_check(NULL, NULL);
+        /* gtk_init_check is called from telega_appindicator_loop so
+         * GTK initialization and the GMainLoop driving its events run
+         * on the same thread.  GTK is documented as not thread-safe
+         * and must be used from the thread that called gtk_init.
+         * Splitting the two caused idle callbacks scheduled via
+         * g_idle_add from telega_appindicator_send to never be
+         * dispatched. */
+        return true;
```
zevlg: Better to remove this function, its declaration and use in the code at all

zevlg (changes requested):

zevlg:
Merged in, thank you!

server/telega-appindicator.c
asgardthundee: <nice>
 /<work>\

PR #572, opened 2026-04-30
fix(contrib): handle matrix media placeholder captions

LuciusChen:
Recognize bridge-generated media captions like `Alice sent an image` so bridge messages keep the real sender and drop the duplicated placeholder caption.

Also make the shared sender helper return the sender id for forwarded messages and stop guessing captionless media senders from recent cache.

## Before

Snipaste_2026-04-30_23-05-00

## After

image

## Matrix Source (Element)

image

Reference: https://t.me/ArchlinuxCN_Appearance/294784


zevlg:
Merged in, thanks!

PR #573, opened 2026-05-10
Ignore missing chats in last message updates

benthamite:
## Summary

- Avoid asserting when `updateChatLastMessage` references a chat that is not present in the local cache.
- Skip the update in that case instead of trying to mutate and mark a nil chat.

## Rationale

TDLib can emit chat update events for chats that are not currently known locally. In that case `telega-chat-get` with the `offline` flag returns nil, and the existing `cl-assert` aborts update processing. Treating the missing chat as a no-op keeps the event handler robust while preserving the existing behavior for cached chats.


zevlg:
No. This assert is correct. If chat is not present in the cache it means telega, TDLib or you local custom code has error. Please collect TDLib logs before you get this assertion and send them privately to me @zevlg in the Telegram

PR #578, opened 2026-07-06
Add missing reaction emoji aliases

roife:
Add missing emoji aliases used by Telegram reactions to `etc/emojis.alist`, including:

  - `🤯`, `🥰`, `🤬`, `🤩`, `🤮`, `🥱`, `🤨`, `🫡`, `🤪`
  - `❤`, `❤‍🔥`
  - `🗿`, `🕊`, `🌭`, `☃`, `✍`
  - `👨‍💻`, `🤷‍♂`, `🤷‍♀`


roife:
This PR makes it more convenient to choose reactions.

Before: 
Clipboard_Screenshot_1783318411

After:
Clipboard_Screenshot_1783318439

zevlg:
Adding emojis that are in the standard is ok.  Also, change for `:shrug-woman:` to use gendered version is ok.  The problem with non-emojis symbols such as: ❤, ❤‍🔥, 🕊, 🌭, ☃, ✍, 🤷‍♂, 🤷‍♀ .  Instead of adding them into emojis.alist it is better to add fe0f workaround into `telega-msg-reaction-title-for-completion` and commenting out that this is done because of incorrect implementation for emoji standard in the Android/iOS keyboards.

Something like this in the `telega-msg-reaction-title-for-completion`:
```emacs-lisp
  ...
(when-let ((emoji-name (or (telega-emoji-name emoji)
                           ;; NOTE: Workaround for non-standard
                           ;; emojis used by Android/iOS keyboards
                           ;; ref. https://t.me/tdlibchat/42441
                           (telega-emoji-name (concat emoji "\xfe0f")))))
  ...
```

zevlg:
Merged in,  thank you for the contribution!

PR #579, opened 2026-07-10
Fix byte/native compilation of `telega-transient.el`

ska80:
During compilation, the helper `telega-transient--keymap-prefix-name` is not yet available to the macro expander (in `telega-transient-define-prefix-by-keymap` macro), resulting in error:

```
Symbol’s function definition is void: telega-transient--keymap-prefix-name
```

Wrapping the `telega-transient--keymap-prefix-name` in `eval-and-compile` ensures it is available when the macro is expanded, fixing the error.

### DIFF
--- a/telega-transient.el
+++ b/telega-transient.el
@@ -1873,8 +1873,9 @@ Return first applicable imc."
   :type '(repeat symbol)
   :group 'telega-modes)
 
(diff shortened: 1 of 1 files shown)


zevlg:
Merged in, thank you!

PR #581, opened 2026-07-17
Fix chat completion fallback after sticker lookup

LuciusChen:
telega-chatbuf-complete-sticker-by-emoji returned t even when the input was not an emoji. Since it runs before completion-at-point, this stopped the completion chain and broke TAB completion for mentions such as @ and @@.

Return t only when sticker completion actually handles an emoji.

Test: make test_el (19/19 passed).


zevlg:
Merged in, thank you!

PR #582, opened 2026-07-21
Add bulk actions for chat join requests

LuciusChen:
## Summary

- make the pending join-request footer entry one actionable button, so both mouse clicks and `RET` open the join-request view
- add **Approve All** and **Decline All** actions with confirmation
- add a thin wrapper for TDLib's `processChatJoinRequests`
- add English and Chinese UI strings

Each bulk action invokes `processChatJoinRequests` once, following TDLib's documented all-pending-requests semantics, and refreshes the join-request view when the call completes.

## Tests

- `make test` — 19/19 Emacs tests and 23/23 server tests pass
- verified with a mocked callback that one bulk action makes exactly one `processChatJoinRequests` call and no follow-up query
- `git diff --check` and plist/parenthesis validation pass

### DIFF
--- a/etc/langs/en.plist
+++ b/etc/langs/en.plist
@@ -1205,8 +1205,6 @@ Example: 23 y.o. designer from San Francisco")
   :value "Approve All")
  ("telega_group_requests_decline_all"
   :value "Decline All")
- ("telega_group_requests_processing"
-  :value "Processing…")
  ("telega_query_group_requests_approve_all"
   :one_value "Approve all {count} request to join «{title}»?"
   :other_value "Approve all {count} requests to join «{title}»?")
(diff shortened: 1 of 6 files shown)


zevlg:
`getChatJoinRequests` is supposed to process all pending join requests without limit.  Probably, during processing, new join requests are made and they are left behind?  In any way, this should be fixed on Telegram server side, not on client side.

Could you please change PR to call `getChatJoinRequests` only once, without loop, to accept/dismiss all pending join requests

LuciusChen:
Updated in d1e45e2. Bulk actions now call `processChatJoinRequests` exactly once and refresh the join-request view when it completes. I removed the retry loop, the follow-up `getChatJoinRequests` query, and the in-progress state. Thanks for the clarification.

zevlg:
Merged in, thank you for the contribution!

PR #583, opened 2026-07-25
Fix sending stickers with TDLib 1.8.65

LuciusChen:
TDLib 1.8.65 moved sticker media fields into the nested `inputSticker` object, matching the new wrappers for other media. telega still constructed the legacy flat `inputMessageSticker`, so `sendMessage` returned an error and the sticker disappeared from the input without being sent.

- Construct the nested `inputSticker` and pass `emoji` at the `inputMessageSticker` level.
- Teach input-file extraction and the one-line input preview about the nested structure.

Verified with a targeted ERT regression covering the generated types, file and thumbnail IDs, dimensions, emoji, and input-file extraction. `git diff --check` also passes.

TDLib schema: https://github.com/tdlib/td/blob/master/td/generate/scheme/td_api.tl#L5024-L5029

### DIFF
--- a/telega-chat.el
+++ b/telega-chat.el
@@ -4961,6 +4961,8 @@ Return valid \"messageSendOptions\"."
      (telega--tl-get imc :audio :audio))
     (inputMessageAnimation
      (telega--tl-get imc :animation :animation))
+    (inputMessageSticker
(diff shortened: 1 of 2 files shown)


zevlg:
Merged in, thank you!

PR #584, opened 2026-07-26
Scale box button brackets to the content height

LuciusChen:
Box button brackets stay too short when emoji, CJK glyphs, embedded images or taller fonts increase the line height.

Rebased onto upstream master now that #613 has merged; the diff contains only this change.

Add `:height content` to measure the styled line prefix and button body once, then share the height and baseline between both SVG brackets. Buttons in the same rich-message or ordinary keyboard row use the tallest label's metrics. Keep outlines aligned with the body and preserve transparent areas during highlighting.

Text-only rendering, unavailable metrics and multiline labels retain the fixed-size fallback. Mixed text and image brackets are supported.

Validation: 38 ERT tests pass on Emacs 29.4, 30.2 and master, including mixed-height rich-message and ordinary keyboard rows. GUI checks cover CJK, emoji, font scaling and multiline fallback.

<details>
<summary>Before / after screenshots</summary>

### Before

image

image

image

image

image

image

### After

image

image

image

image

image

image

</details>

### DIFF
--- a/telega-core.el
+++ b/telega-core.el
@@ -2147,14 +2147,30 @@ Return what BODY returns."
   (declare (indent 1))
   (let ((style-sym (gensym "style"))
         (left-bracket-sym (gensym "left-bracket"))
-        (right-bracket-sym (gensym "right-bracket")))
-    `(let ((,style-sym ,style))
+        (right-bracket-sym (gensym "right-bracket"))
+        (content-height-p-sym (gensym "content-height-p"))
+        (left-start-sym (gensym "left-start"))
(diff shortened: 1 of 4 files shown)


zevlg:
This looks very promissing! let me first check this locally before merging it in

LuciusChen:
I've tested all the scenarios shown in the screenshots above, and they now work as expected.

LuciusChen:
Added an explicit :background nil to SVG bracket images so their transparent area can follow cursor-face highlighting instead of retaining the passive background. The full test suite still passes (26/26).

PR #585, opened 2026-07-27
Fix Instant View with TDLib 1.8.66

LuciusChen:
## Summary

Update the Instant View renderer for TDLib 1.8.66 schema changes.

- Read `blocks` instead of the removed `page_blocks` field from `webPageInstantView`, list items, collages, slideshows, and details blocks.
- Use the TDLib `is_open` state when rendering and toggling details blocks.
- Handle the new `richTextReference` target and `richTextReferenceLink` link types as in-page anchors.
- Restore automatic button property retention in `telega-button--insert`, matching its documented behavior and the implementation before 6ff8203. Without it, updating a details button drops its inserter/action and the first toggle fails with `Button :inserter is unset`.

TDLib changes:

- [Simplify PageBlock field names](https://github.com/tdlib/td/commit/4f3676d76ccc082354b5b0a5645cecad947dfa04)
- [Add td_api::richTextReferenceLink](https://github.com/tdlib/td/commit/d17125721bbaf95fe4dd5801f1837f3acf9e83d0)

## Root cause

The top-level `webPageInstantView` object now returns its content in `:blocks`, while `telega-webpage--instant-view` still looked up `:page_blocks`. The lookup therefore returned nil and produced an empty Instant View buffer. The same rename affected several nested PageBlock types.

## Tests

Added ERT coverage for:

- top-level Instant View blocks
- nested list, collage, and slideshow blocks
- repeated details close/reopen
- reference-link navigation

Full local suite: **23 tests passed, 0 unexpected**.

### DIFF
--- a/telega-core.el
+++ b/telega-core.el
@@ -1618,6 +1618,8 @@ Properties specified in PROPS are retained on
           (apply #'make-text-button beg end
                  :type button-type
                  :value value
+                 :telega-retain-props
+                 (telega-plist-map (lambda (prop _ignored) prop) props)
                  props)))
     (when-let ((add-sensor (plist-get props :telega-add-sensor-func)))
       (telega--change-text-property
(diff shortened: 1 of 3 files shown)


zevlg:
Thank you, merged in! IV code needs rework to utilize telega-rich-text.el functionality, this is in TODO

PR #588, opened 2026-08-03
Implement resend-as-file for photos with invalid dimensions

LuciusChen:
Complete the existing `RESEND as file` button for photo messages that fail with `PHOTO_INVALID_DIMENSIONS`.

The action sends the original file as a document, preserves its caption, and removes the failed photo only after TDLib creates the replacement message. Direct pre-send errors use the existing error handling.

Validation: 25 ERT tests passed, including button activation, emoji captions, and rejected resend requests with mocked TDLib calls. Byte compilation and `check-parens` passed on Emacs 32.0.50.

### DIFF
--- a/telega-chat.el
+++ b/telega-chat.el
@@ -3042,7 +3042,10 @@ If NEW-FOCUS-STATE is specified, then focus state is forced."
 
 (defun telega-chatbuf-msg--pp-inserter (msg)
   "Return message inserter for the pretty printer."
-  (cond ((telega-msg-match-p msg
+  (cond ((plist-get msg :telega-resend-as-file)
+         #'telega-ins--msg-sending-state-failed)
+
+        ((telega-msg-match-p msg
            `(and is-deleted
                  (chat ,telega-chat-show-deleted-messages-for)))
          #'telega-ins--message-deleted)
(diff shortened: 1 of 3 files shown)


LuciusChen:
Updated the PR to cover the 10 MiB TDLib photo limit reported in https://t.me/emacs_telega/51487. Oversized ordinary photos now use the same document fallback as photos with invalid dimensions.

zevlg:
We can't do this.  Limits changes and not documented.  We should display [Resend as file] if photo message sending fails due to limits.

LuciusChen:
Updated as suggested: removed the undocumented size/dimension pre-checks and restored the normal photo send path. The PR now shows `RESEND as file` only after TDLib rejects the photo, including direct `sendMessage` errors that don't create a failed message.

zevlg:
I'm not sure about this, too complicated.  Let's hold on this

I think we should display [resend as file] only if Telegram server reports that message can't be sent.  If TDLib reports error before message is sent to the server telega should report error and keep user input as is

LuciusChen:
That makes sense. The extra complexity came from the oversized-photo case: TDLib returns an error directly from sendMessage without creating a failed message, so I added a local failure notice to expose the action. Keeping the original input unchanged for such pre-send errors is a cleaner model. Let's hold this for now.

PR #589, opened 2026-08-04
Fix draft messages with TDLib 1.8.66

LuciusChen:
## Summary

- save and restore drafts with both draftMessage representations found in TDLib 1.8.66 snapshots
- replace the chat-list draft placeholder with the actual draft content
- support text, rich-message, voice-note, and video-note draft previews

Some TDLib 1.8.66 snapshots use draftMessage.content while later ones use draftMessage.input_message_text. Sending and reading both forms keeps telega compatible with either schema.

## Testing

- make compile with Emacs 32.0.50
- make test_el: 23/23 tests passed
- verified incomplete rich-message drafts do not create a button without a message target

### DIFF
--- a/telega-chat.el
+++ b/telega-chat.el
@@ -3541,7 +3541,7 @@ otherwise set draft only if chatbuf input is also draft."
          (goto-char telega-chatbuf--input-marker)
          (telega-ins--with-props '(:draft-input-p t)
            (telega-ins--fmt-text
-            (telega--tl-get draft-msg :input_message_text :text))))))))
+            (telega--tl-get draft-msg :content :text))))))))
 
 (defun telega-chatbuf--load-initial-history ()
   "Load initial history in the chatbuf."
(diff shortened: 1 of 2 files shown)


zevlg:
Keeping backward compatibility is a bad idea.  Backward compatibility can't be done nicely in general, it makes code more complex and harder to maintain.

Please create dedicated inserter function for the draft message content instead of doing this directly in the `telega-ins--input-content-one-line`,  in that function you can create "fake" imcs to utilize `telega-ins--input-content-one-line` in order to avoid code duplication.  Draft message content is another structure and its field names won't  match with imc field names in some future

LuciusChen:
Updated, thanks. The PR now targets only draftMessage.content, and draft content rendering uses a dedicated inserter that converts supported draft types to fake IMCs before reusing telega-ins--input-content-one-line.

zevlg:
Very nice! Merged in, thank you

I'll make some modifications to the code on next commit to make it a little more maintable

PR #591, opened 2026-08-06
Fix quoting hidden spoilers

LuciusChen:
Hidden spoilers are rendered as block characters in the chat buffer, so constructing inputTextQuote from the selected buffer text sends blocks instead of the original message text.

Build the quote from the message formattedText using UTF-16 region offsets. This preserves spoiler entities while keeping the quote preview hidden.

Tests: all 23 ERT tests pass, along with a focused hidden-spoiler and emoji-offset regression.

### DIFF
--- a/telega-chat.el
+++ b/telega-chat.el
@@ -6564,31 +6564,37 @@ Return non-nil if message MSG has been redisplayed."
 
 (defun telega-chatbuf--input-text-quote ()
   "Return TL inputTextQuote from currently selected region."
-  (when (and (region-active-p)
-             (when-let ((msg (telega-msg-at)))
-               ;; 1. Check message has some text to work with
-               ;; 2. Check region spans on the same message
-               (and (telega-msg-content-text msg)
-                    (eq msg (telega-msg-at (region-beginning)))
-                    (eq msg (telega-msg-at (region-end))))))
-    ;; NOTE: Check reply quote limits first, to avoid "400: Message is
-    ;; too long" errors
-    (when-let ((quote-limit
-                (plist-get telega--options :message_reply_quote_length_max)))
-      (when (> (- (region-end) (region-beginning)) quote-limit)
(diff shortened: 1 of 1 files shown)


telega-chat.el
```diff
@@ -6564,31 +6564,37 @@ Return non-nil if message MSG has been redisplayed."
...
-                     (buffer-substring (region-beginning) (region-end)))
-              :position (if-let ((content-begin
-                                  (save-excursion
-                                    (telega-chatbuf--goto-msg-content 0))))
-                            (- (region-beginning) content-begin)
-                          0))
-      (deactivate-mark))))
+  (when-let* (((region-active-p))
+              (msg (telega-msg-at))
+              (msg-fmt-text (or (telega--tl-get msg :content :text)
```
zevlg: Could you please name this var as `raw-msg-content-text`, so if `telega-msg-content-text` changes its logic `msg-content-text` will be grepable

telega-chat.el
```diff
@@ -6564,31 +6564,37 @@ Return non-nil if message MSG has been redisplayed."
...
+                                (telega--tl-get msg :content :caption)))
+              ((eq msg (telega-msg-at (region-beginning))))
+              ((eq msg (telega-msg-at (region-end))))
+              (content-begin
+               (save-excursion
+                 (telega-chatbuf--goto-msg-content 0))))
+    (let* ((quote-position
```
zevlg: Maybe put this let vars under `when-let*` as they are non-nil for sure?

zevlg (changes requested):

telega-chat.el
LuciusChen: Renamed it to raw-msg-content-text.

telega-chat.el
LuciusChen: Moved both bindings into the surrounding when-let*.

zevlg:
Merged in, thank you!

PR #593, opened 2026-08-06
Fix sending voice and video notes with TDLib 1.8.66

OSadovy:
## Summary

TDLib moved voice and video note media fields into nested `inputVoiceNote` and `inputVideoNote` objects -- the same wrapping #583 handled for stickers. telega still constructed the legacy flat form, so `sendMessage` returned an error and the note disappeared from the input without being sent.

- Construct the nested `inputVoiceNote` / `inputVideoNote`, moving `duration`, `waveform`, `length` and `thumbnail` inside.
- Teach input-file extraction and the one-line input preview about the nested structure.

TDLib changes:

- [Add td_api::inputVoiceNote](https://github.com/tdlib/td/commit/bfc292bd52a1ef248b1a4873b935a0bf090e0a96)
- [Add td_api::inputVideoNote](https://github.com/tdlib/td/commit/2caa367404282001c32a9d72911a8d1a60805db1)

## Root cause

`telega-chatbuf-attach-voice-note` built `inputMessageVoiceNote` with `voice_note`, `duration` and `waveform` at the top level. TDLib reads the single `voice_note` field as an `inputVoiceNote`, so the `inputFileLocal` was parsed as the wrapper -- no file inside, duration 0, empty waveform -- and rejected it:

```
error { code = 400 message = "InputFile is not specified" }
```

The rejection is silent. `sendMessage` is synchronous unless `telega-chat-send-messages-async` is set, so the error is returned rather than delivered through the event loop, while the input is cleared either way. Only the preview path, which passes a callback and is therefore asynchronous, surfaced it at all.

Two call sites move with the construction:

- `telega-chatbuf--input-imc-file` must descend one level further, as it already does for the types wrapped earlier, or the preliminary upload started for the attachment is never cancelled.
- `telega-ins--input-content-one-line` prefers the wrapper and falls back to the content itself, because `draftMessageContentVoiceNote` / `draftMessageContentVideoNote` keep these fields flat and reach the same renderer as a fake imc.

## Tests

Added ERT coverage:

- input-file extraction from wrapped voice and video note content
- one-line input preview for both the wrapped attachment and the flat draft shape

Both new tests fail on `master` and pass with this change.

- `make compile` with Emacs 30.2: clean, no new warnings
- `make test_el`: **25 tests passed, 0 unexpected**

Voice notes verified end to end against TDLib 1.8.66 -- recorded, sent, and played back. Video notes are not exercised; the schema change is the same one and the affected code paths are shared.

### DIFF
--- a/telega-chat.el
+++ b/telega-chat.el
@@ -4964,9 +4964,9 @@ Return valid \"messageSendOptions\"."
     (inputMessageSticker
      (telega--tl-get imc :sticker :sticker))
     (inputMessageVideoNote
-     (plist-get imc :video_note))
+     (telega--tl-get imc :video_note :video_note))
     (inputMessageVoiceNote
-     (plist-get imc :voice_note))))
+     (telega--tl-get imc :voice_note :voice_note))))
 
 (defun telega-chatbuf-input-send (arg &optional preview-p)
   "Send chatbuf input to the chat.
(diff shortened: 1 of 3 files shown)


zevlg:
Thank you for adding support for TDLib's 1.8.66 video/voice notes.  I'll remove fallback to IMC itself to keep support for fake imcs in draft messages.  draft messages should generete correct fake imc of these kinds

PR #594, opened 2026-08-10
Make monochrome SVG symbols follow face foreground

LuciusChen:
SVG uses black as the default fill, so older monochrome symbols stay black when displayed with a light foreground in dark themes. Use currentColor for their inherited fill, matching newer symbol assets and allowing Emacs to use the current face foreground.\n\nThis also fixes mixed-color symbols whose outlines already used currentColor while their filled paths still defaulted to black.\n\nTests:\n- make test_el (23 tests)\n- xmllint --noout etc/symbols/*.svg


zevlg:
Are you sure these svg images does not inherit face's foreground?  Emacs wraps svg images to have currentColor as fill color, so if there is no explicit fill color inside svg image, it should inherit face's foreground.

For example:
screenshotF2BPA0

LuciusChen:
This changed after Emacs 30 and is already present in the emacs-31.0.90 tag. I reproduced it with -Q on the same macOS system: Emacs 30.2 inherits the face foreground, while Emacs 32.0.50 renders the original SVG in black.

Emacs commit [4d145d08](https://github.com/emacs-mirror/emacs/commit/4d145d08da86a05187f99ab6352637241e146547) removed fill: currentColor from the SVG wrapper and updated Emacs’s bundled monochrome SVGs to declare it explicitly. The wrapper now only sets color, which does not set fill.

So this PR applies the same migration to telega’s monochrome SVGs and remains compatible with Emacs 30.

zevlg:
Ah, I see

zevlg:
Merged in, thank you!

PR #595, opened 2026-08-10
Fix CAPF search bounds outside chat input

yibie:
## Summary

- Return nil from emoji completion when its backward-search bound is after point.
- Return nil from bot-command completion when point is before the chat input marker.
- Add a regression test for both public CAPFs outside the chat input.

## Problem

With automatic CAPF frontends such as Corfu, completion can be requested while point is in the chat history, for example around an asynchronous chat-buffer redraw. The emoji and bot-command bounds helpers then pass a search bound to `re-search-backward` that is on the wrong side of point, which signals:

```text
Invalid search bound (wrong side of point)
```

Outside the input area these CAPFs should simply return nil.

## Verification

- `make test_el` — 24/24 ERT tests passed.
- Focused regression reproduced the error before the fix and passed afterward.
- `telega-completions.el` byte compilation and `check-parens` passed.

### DIFF
--- a/telega-completions.el
+++ b/telega-completions.el
@@ -505,6 +505,7 @@ Handles repeated leading CHARs (e.g. @@).  Returns nil if not applicable."
 (defun telega-capf--botcmd-bounds ()
   "Return (START . END) if point is in a /command at start of chatbuf input."
   (when (and telega-chatbuf--input-marker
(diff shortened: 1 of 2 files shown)


zevlg:
Merged in, thank you!

yibie:
So happy to know that I can help.

PR #597, opened 2026-08-11
Generate correct fake imcs for voice and video note drafts

OSadovy:
## Summary

Follow-up to #593, doing what you described: drafts now generate a correct fake imc, so the fallback to the content itself can go.

- `telega-ins--draft-content-one-line` nests a note's `file_path`, `duration`, `waveform` and `length` into `inputVoiceNote` / `inputVideoNote` instead of relabelling the draft's `:@type` and passing the flat fields through.
- `telega-ins--input-content-one-line` now reads only the wrapper for both note kinds.

Relabelling also never produced an `InputFile` -- a draft carries `file_path`, a plain string -- so the wrapper is built with the `inputFileLocal` it should have had.

## Tests

`telega-vvnote-tdlib-1.8.66-one-line` now drives the draft cases through `telega-ins--draft-content-one-line` with real `draftMessageContentVoiceNote` / `draftMessageContentVideoNote`, rather than the hand-built flat imc it used before -- so it covers the conversion rather than the tolerance that replaced it.

Verified the coverage is real: with the renderer made strict but the old relabelling left in place, the draft cases fail.

- `make compile` with Emacs 30.2: clean, no new warnings
- `make test_el`: **26 tests passed, 0 unexpected**

### DIFF
--- a/telega-ins.el
+++ b/telega-ins.el
@@ -4211,11 +4211,7 @@ If SHORT-P is non-nil then use short version."
                     (telega-ins--self-destruct-type tl-ttl 'short)))
                  ))))
      (inputMessageVoiceNote
-      ;; The duration and waveform sit inside the `inputVoiceNote' this is
-      ;; built with, but a draft keeps them flat -- `draftMessageContentVoiceNote'
-      ;; has no wrapper -- and drafts reach this same renderer as a fake imc.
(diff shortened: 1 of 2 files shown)


zevlg:
Merged in, thank you! I'll make a little modification to this code to make it more maintable

PR #599, opened 2026-08-16
Sum filter button counts without building a list per button

OSadovy:
I came to this while profiling redisplay slowdowns in the root buffer. It turned
out not to be the cause of what I was chasing, but it is a small win on its own
and it is already written — so I would rather send it than leave it on a shelf.

`telega-ins--filter` runs for every custom filter and folder button, and a rootbuf
redisplay follows each batch of updates — so this sits between one keystroke and
the next. Summing through `mapcar` builds two throwaway lists the size of the
folder every time, which for a folder holding most of an account is several
hundred cons cells per button per redisplay.

`cl-loop` sums in place and allocates nothing.

Byte-compiled, 2000 calls of both counts, on chats carrying a full complement of
TDLib fields (Emacs 30.2):

| chats in folder | `mapcar` + `apply` | `cl-loop` |
|---|---|---|
| 100 | 22.3 µs/button, 0 GC | 16.6 µs/button, 0 GC |
| 500 | 134.7 µs/button, 2 GC | 98.9 µs/button, 0 GC |
| 1500 | 384.7 µs/button, 6 GC | 261.4 µs/button, 0 GC |

Roughly 1.4× faster, and it takes the summation off the garbage collector
entirely — which is the part that shows up as a pause rather than as slowness.

No behaviour change: same values, same order, `cl-loop ... sum` yields 0 for an
empty folder exactly as `(apply #'+ nil)` did.

No test — this is a pure refactor of an arithmetic expression with no observable
output change. Happy to add one if you would like.

### DIFF
--- a/telega-filter.el
+++ b/telega-filter.el
@@ -138,9 +138,12 @@ See `telega-filter--ewoc-spec' for CUSTOM-SPEC description."
          ;; NOTE: Folders always active
          (active-p (or (telega-filter--folder-p filter) (not (null chats))))
          (nchats (length chats))
-         (unread (apply #'+ (mapcar (telega--tl-prop :unread_count) chats)))
-         (mentions (apply #'+ (mapcar
-                               (telega--tl-prop :unread_mention_count) chats)))
+         ;; NOTE: sum in place, `mapcar' allocated a list per button per
+         ;; redisplay.  See https://github.com/zevlg/telega.el/pull/599
+         (unread (cl-loop for chat in chats
+                          sum (plist-get chat :unread_count)))
+         (mentions (cl-loop for chat in chats
+                            sum (plist-get chat :unread_mention_count)))
          (umstring
           (telega-ins--as-string
            (telega-ins--with-attrs (list :max 7


telega-filter.el
```diff
@@ -138,9 +138,10 @@ See `telega-filter--ewoc-spec' for CUSTOM-SPEC description."
          ;; NOTE: Folders always active
          (active-p (or (telega-filter--folder-p filter) (not (null chats))))
          (nchats (length chats))
-         (unread (apply #'+ (mapcar (telega--tl-prop :unread_count) chats)))
-         (mentions (apply #'+ (mapcar
-                               (telega--tl-prop :unread_mention_count) chats)))
+         (unread (cl-loop for chat in chats
```
zevlg, line 143: Could you please add comment with the link to https://github.com/zevlg/telega.el/pull/599

zevlg (changes requested):

OSadovy:
Added, thanks.

zevlg:
Merged in, thank you!

PR #606, opened 2026-09-06
Fix attaching a file already uploaded to TDLib

rayslava:
Attaching a video from telega's cache signals
`wrong-type-argument stringp nil'.

`telega-file--upload' returns nil when the file is already uploaded and
no `:update-callback' is given, so `telega-chatbuf--gen-input-file'
evaluates `(telega--tl-error-p nil)' -> `(intern nil)'.

Second commit fixes an unrelated typo: the `inputVideo' object in
`telega-chatbuf-attach-video' uses `:type' instead of `:@type'.


zevlg:
> `telega-file--upload' returns nil when the file is already uploaded and no `:update-callback' is given

So we need to fix `telega-file--upload` instead.  It should return file object as described in its docstring.  We have parents misalignment there, and `ufile` should be after `if` clause not inside its `else` part

telega-media.el
```diff
@@ -272,8 +272,8 @@ Return file object, obtained from `telega--preliminaryUploadFile'."
         (telega-file--add-update-callback (plist-get ufile :id)
           (lambda (file)
             (funcall update-callback file)
-            (telega-file--uploading-p file))))
```
zevlg, line 275: Fix `telega-file--upload` instead, it should always return file object

zevlg (changes requested):

telega-media.el
rayslava: Updated

Issue #605, opened 2026-09-06
593 native-comp warnings from missing declare-function forms


anevolbap:
## Telega Setup
*OS*: Linux 6.12.95+deb12-amd64 (Debian), x86_64
*Emacs*: GNU Emacs 31.0.60 (build 1, x86_64-pc-linux-gnu, GTK+ 3.24.38)
*Telega*: 0.8.660 (git, 79c1735)

Installation method: git

## Current Behavior

Async native compilation reports 593 "is not known to be defined" warnings, spread over 25 of the 36 `telega*.el` files. Each file compiles in a pristine subprocess, so any cross-file call with no `require` and no `declare-function` goes unresolved. `telega-ins.el` alone accounts for 122.

The mechanism is already in place: the tree carries 144 `declare-function` forms. It covers a fraction of the cross-file calls, and 593 file-and-function pairs still lack one.

`require` cannot cover these. `telega-chat.el:65` requires `telega-ins`, while `telega-ins.el:4981` and `:5091` call `telega-chat-member-my-status` and `telega-chat-private-p` from `telega-chat.el`. The graph is circular.

These are warnings, not errors. Nothing fails at runtime.

## Steps to Reproduce

Compile each file in its own process, which is what async native comp does:

```elisp
;; one.el
(add-to-list 'load-path "/path/to/telega.el")
(byte-compile-file (car (last command-line-args)))
```

```console
$ for f in telega*.el; do emacs --batch -l one.el "$f"; done 2>&1 \
    | grep -c "is not known to be defined"
593
```

That leaves `.elc` files next to the sources, worth removing afterwards.

A single batch process reports only 228, because each `require` warms the environment for the files compiled after it.

## Possible Solution

Inserting the missing `declare-function` forms after `;;; Code:` in each file takes the count from 593 to 5. Comparing every warning category before and after, none is new and none increased. Per file, `native-compile` drops from 5 to 0 on `telega-community.el`, 36 to 0 on `telega-topic.el`, 18 to 0 on `telega-voip.el`, and 36 to 0 on `telega-emoji.el`.

The 5 that remain are not declaration problems:

- `telega-filter.el` calls `ewoc-enter-last` and `ewoc-collect` with no `(require 'ewoc)`
- `telega-transient.el` calls `ewoc-delete` with no `(require 'ewoc)`
- `telega-rich-text.el` calls `org-create-formula-image` with no `(require 'org)`
[...]

PR #607, opened 2026-09-06
Declare cross-file functions so each file compiles on its own

anevolbap:
Addresses the native compilation warnings reported in #605. Reworked after review.

Async native compilation starts from a pristine environment and compiles each file in its own subprocess, so every cross-file call with no `require` and no `declare-function` is reported as unknown. Compiling the tree one file per process gives 620 such warnings across 25 of the 36 files.

`make compile` shows none of this, since it compiles everything in one Emacs where the requires have already warmed the environment. Reproducing it needs one process per file, or async native compilation as it runs in a normal session.

Six commits, in the order suggested by review.

**Requires first.** 68 requires across 21 files, everywhere the required module does not already reach back to the requiring one. Every candidate edge was tested against the transitive closure, added one at a time, since edges that are each individually safe can jointly close a loop. This covers the cases raised in review: `telega-emoji` requires `telega-ins`, `telega-media`, `telega-modes`, `telega-sticker`, `telega-tdlib` and `telega-util`; `telega-filter` requires `telega-i18n` and `telega-util`; `telega-vvnote` requires `telega-ins`.

**Then the three moves and two small fixes.** `telega-plist-del`, `telega-plist-map` and `telega-focus-state` move from `telega-util` to `telega-core`, which uses them. The `telega-avatar-text-function` default is plain-quoted rather than sharp-quoted, since a defcustom default is a symbol. And `telega-string-width` is wrapped in `eval-and-compile`: both branches of a bare top-level `if` are invisible to the compiler, which is why it read as undefined at its own call sites, and why declaring it was wrong.

**Declarations last, for what is genuinely cyclic.** 335 across 21 files. Every remaining module edge was checked; none can become a require without a cycle. Each goes into the file's existing declaration block rather than a second one, and carries the arglist copied from the source definition, so `check-declare` reports exactly the 17 warnings it did before this branch.

Result: 620 warnings to 4. `make compile` and `make test_el` pass on Emacs 31.0.60 and 28.2, 28 tests, 0 unexpected.

Adding the requires also fixed three real miscompiles as a side effect. `with-telega-help-win` (`telega-core.el:1050`) expands to a `telega-server--callback-put` call, and where `telega-server` was not loaded the macro call stayed literal and compiled into a call to a function that does not exist. `telega-emoji`, `telega-story` and `telega-topic` now expand it to `(puthash ...)` correctly.

## The 4 remaining, left deliberately

`telega-core.el` still hits the same `telega-server--callback-put` problem, and cannot be fixed the same way because `telega-server.el` requires `telega-core`. Declaring it would silence the warning and leave the miscompile.

`telega-tdlib.el:3070` does `(setf (telega-story-deleted-p reply) t)`, whose `gv-define-setter` lives in `telega-story.el`, which `telega-tdlib.el` does not require:

```elisp
;; without telega-story loaded
(let* ((v reply)) (\(setf\ telega-story-deleted-p\) t v))
;; with it
(let* ((v reply)) (plist-put v :telega-is-deleted-story t))
```

`(setf telega-story-deleted-p)` is not `fboundp`. That path runs when `telega--getStory` gets a TL error, the deleted or expired story case it exists to handle.

Both need a layering decision, so they are left for #605.

The last two are `tracking-add-buffer` and `tracking-remove-buffer` in `telega-tdlib-events.el`, from the optional `tracking` package, which is outside this change.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)


telega-core.el
```diff
@@ -41,6 +41,28 @@
 (declare-function telega-em-width-ratio "telega-util")
 (declare-function telega-em-height-ratio "telega-util")
 
+;; Forward declarations, so each file compiles cleanly on its own.
+(declare-function telega-string-width "telega-core")
```
zevlg: Why declaring function that is defined below?

telega-core.el
```diff
@@ -41,6 +41,28 @@
...
+(declare-function telega-custom-emoji-get "telega-emoji")
+(declare-function telega-folder-name "telega-folders")
+(declare-function telega--MessageTopic "telega-tdlib")
+(declare-function telega--getCustomEmojiStickers "telega-tdlib")
+(declare-function telega--change-text-property "telega-util")
+(declare-function telega--fmt-text-faces "telega-util")
+(declare-function telega--split-by-text-prop "telega-util")
+(declare-function telega-color-name-set-saturation-light "telega-util")
+(declare-function telega-duration-human-readable "telega-util")
+(declare-function telega-focus-state "telega-util")
```
zevlg: Probably we should move `telega-focus-state` to the telega-core itself

telega-core.el
```diff
@@ -41,6 +41,28 @@
...
+(declare-function telega-plist-del "telega-util")
+(declare-function telega-plist-map "telega-util")
```
zevlg: Same here, we should move `telega-plist-del` and `telega-plist-map` to the telega-core.el

telega-customize.el
```diff
@@ -32,6 +32,9 @@
 
 (require 'org-faces)
 
+;; Forward declarations, so each file compiles cleanly on its own.
+(declare-function telega-avatar-text-simple "telega-media")
```
zevlg: Better to use `'` instead of `#'` for `telega-avatar-text-function` value

telega-emoji.el
```diff
@@ -31,6 +31,43 @@
 (declare-function telega-chat--update "telega-tdlib-events" (chat &rest events))
 (declare-function telega-match-p "telega-match-p" (object temex))
 
+;; Forward declarations, so each file compiles cleanly on its own.
```
zevlg: `telega-ins`, `telega-media`, `telega-modes`, `telega-sticker`, `telega-tdlib` and `telega-util` can be required from `telega-emoji.el` without making require loops.  So declarations won't be needed

telega-filter.el
```diff
@@ -71,6 +72,18 @@
 (declare-function telega-root--buffer "telega-root") ; `with-telega-root-buffer'
 (declare-function telega-root-view--redisplay "telega-root")
 
+;; Forward declarations, so each file compiles cleanly on its own.
```
zevlg: You can require `telega-i18n` and `telega-util` without creating require loops

telega-vvnote.el
```diff
@@ -29,6 +29,13 @@
 (require 'telega-ffplay)
 (require 'telega-util)                  ;`with-telega-symbol-animate'
 
+;; Forward declarations, so each file compiles cleanly on its own.
+(declare-function telega-msg-redisplay "telega-chat")
+(declare-function telega-ins--image "telega-ins")
+(declare-function telega-ins-progress-bar "telega-ins")
```
zevlg: `telega-ins` can be required, `telega-msg` and `telega-chat` can't, so keep declarations from them

zevlg (changes requested):

zevlg:
These declarations are so massive.  It clutters code much and it is nearly impossible to maintain correctness of the declarations.  Probably we should have separate `telega-decls.el`, where all these declarations will live.  Some code reorganization might be required to avoid circular dependencies, so declarations won't be needed at all

anevolbap:
Agreed on all of it. Rebuilding the require graph confirms the specific cases: `telega-emoji` can require `telega-ins`, `telega-media`, `telega-modes`, `telega-sticker`, `telega-tdlib` and `telega-util`; `telega-filter` can require `telega-i18n` and `telega-util`; `telega-vvnote` can require `telega-ins` but not `telega-msg` or `telega-chat`.

That generalises across the tree: 91 module edges are loop-free, which removes 332 of the 585 declarations. The remaining 253 sit on real cycles, concentrated in `telega-ins` (39), `telega-transient` (32) and `telega-media` (27). So `telega-decls.el` would hold roughly a quarter of what is in this PR now.

`telega-string-width` was a plain mistake. It is defined in `telega-core.el` itself, inside a conditional below the use, so it should move above the use rather than be declared. Using `'` instead of `#'` for `telega-avatar-text-function` removes that warning on its own, confirmed. Moving `telega-focus-state`, `telega-plist-del` and `telega-plist-map` into `telega-core` drops more.

Two of the warnings should not become declarations at all, since they are miscompiles rather than missing hints. Details in #605.

Would the preference be one PR doing the requires and the moves first, with `telega-decls.el` for the remainder as a follow-up, or all of it in one go?

anevolbap:
Reworked along those lines, and two numbers above need retracting.

The 593 figure was undercounted. The per-file sweep it came from was running without the third-party packages on `load-path`, so nine file compiles aborted early and their warnings were never counted. With a correct `load-path` master is 620, and seven files' warnings had been missing from the working set entirely.

The split I quoted was also wrong, and in the optimistic direction. 91 loop-free edges was measured by testing each edge against the original graph independently; adding them one at a time, edges that are individually safe jointly close loops. The real figures are 68 requires and 335 declarations remaining, roughly the inverse of what I said.

What is in the branch now, in the suggested order: requires first, then the three moves into `telega-core`, the plain-quoted defcustom default, and `telega-string-width` wrapped in `eval-and-compile` (a bare top-level `if` hides both branches from the compiler, which is why it read as undefined at its own call sites and why declaring it was wrong). Declarations last, only where a require would close a loop, merged into each file's existing block and carrying arglists copied from the source definitions, so `check-declare` reports the same 17 warnings as master.

620 to 4. `make compile` and `make test_el` pass on 31.0.60 and 28.2.

The requires also fixed three real miscompiles on their own. `telega-emoji`, `telega-story` and `telega-topic` were compiling the `telega-server--callback-put` macro call into a call to a function that does not exist, and now expand it to `(puthash ...)`. `telega-core` cannot be fixed that way since `telega-server` requires it back, so that one and the `telega-tdlib` setf are still open, along with two `tracking-*` functions from the optional package.

Still happy to move the remaining 335 into `telega-decls.el` if that is preferred over per-file blocks.

telega-core.el
anevolbap: Right, that was a mistake. The definition sits inside a bare top-level `if`, and the compiler ignores both branches, which is why it read as undefined at its own call sites further down. Wrapped in `eval-and-compile` instead, so the definition is real to the compiler and no declaration is needed.

telega-core.el
anevolbap: Moved. `telega-focus-state` is now defined in telega-core.el and declared nowhere.

telega-core.el
anevolbap: Moved as well. `telega-plist-del` and `telega-plist-map` are in telega-core.el now, both gone from telega-util.el and declared nowhere.

telega-customize.el
anevolbap: Changed to a plain quote. The declaration is gone with it, since a defcustom default is just a symbol and nothing needs resolving at compile time.

telega-emoji.el
anevolbap: Confirmed against the require graph and applied: telega-emoji.el now requires telega-ins, telega-media, telega-modes, telega-sticker, telega-tdlib and telega-util, and has no declarations left.

telega-filter.el
anevolbap: Applied, telega-filter.el requires telega-i18n and telega-util.

telega-vvnote.el
anevolbap: Applied exactly as described: telega-ins is required, and the telega-msg and telega-chat declarations are kept since both reach telega-vvnote back.

anevolbap:
Both approaches are now in the branch so they can be compared directly. `2939d55` has the declarations in per-file blocks; `88ced49` moves them into `telega-decls.el`, and `b71920f` adds a check for them. Happy to drop either commit.

One thing worth knowing before choosing: a `telega-decls.el` that simply contains `declare-function` forms does not work. The byte compiler only honours them in the file being compiled, so requiring such a file has no effect on the requiring one. `byte-compile-macroexpand-declare-function` pushes onto `byte-compile-function-environment`, which is per compilation unit; a `require` or `load` evaluates the forms through the plain `subr.el` macro instead. `require`, `eval-when-compile` and a compile-time `load` all still warn.

So the declarations are stored as data and installed by a macro, which expands inside the compiling file where the compiler does see them:

```elisp
(require 'telega-decls)
(telega-declare-functions telega-ins)
```

Every telega file is then back to the 144 declarations master already has, with the other 335 in one place.

The cost is that `check-declare` only understands literal `declare-function` forms, so it stops covering the 335. Since that is exactly the "impossible to maintain correctness" problem, `b71920f` adds `telega-decls-check`, which walks the table and reports any entry whose function is not defined in the file it names or whose declared arglist disagrees with the source. It runs as part of `make test_el`, so a moved or renamed function fails CI rather than going stale. It reports nothing on the current table, and I checked it does catch a wrong file, a wrong arglist and a missing function.

620 to 4 either way. `make compile` and `make test_el` pass on 31.0.60 and 28.2, 29 tests.

